PR #9, opened 2014-01-29
Service examples

rosscdh:


msabramo:
What is this "Django service"?

I haven't used Django in a while (I mostly use Pyramid these days). 

I was intrigued by the fact that it mentioned Django but when I googled "Django services" I found a bunch of stuff about Django REST Framework but nothing that looked like this. 

It also seems to have no Django dependencies other than the File object. Though maybe it exposes some API that Django knows how to use?

Just curious.

bebraw:
It's an example of how to use the library with Django and the library doesn't have any direct dependency on it. It's just there to show how to integrate with it. There is no reason why there couldn't be a Pyramid example as well.

We could push it below `/examples` to make this clear.

@msabramo Thanks for your PRs btw!

msabramo:
But is this really showing how to use the library with Django specifically? Or is it just generally showing how to use the library? Other than the line:

```python
from django.core.files import File
```

It doesn't seem to be specific to Django. It seems I could use the same code in Pyramid, Flask, web2py, or whatever.

Or am I missing something? Does the fact that it has a `generate` method mean that Django can do something special with it?

bebraw:
> It doesn't seem to be specific to Django. It seems I could use the same code in Pyramid, Flask, web2py, or whatever.

That's actually very valid point. Maybe it would make sense to refactor the classes so that they would receive File as a dependency at init? Then you could use something else instead of Django File if you want. Or, even simpler perhaps, the methods could return `self.file_object` which you would then have to wrap into File etc.

What do you think would be the best approach?

> Does the fact that it has a generate method mean that Django can do something special with it?

I don't think Django does something special with it. It has been a long time since I used Django so it's possible I'm wrong, though.

rosscdh:
Hey there.

Correct there is almost no dependency on django in the objects just the
file, but that can be removed and dep injected or something along those
lines.

Sorry I misnamed the PR. Glad to be able to offer an implementation example.

Best
Ross

Issue #30, opened 2015-01-31
Coordinate a release - 0.90


bebraw:
* [ ] Add @msabramo's contributions to README
* [ ] Figure out what to with #9. @msabramo, I would appreciate your input here.

Am I missing something? I can't think of any other tasks.

msabramo:
I like the idea of renaming `services.py` and putting it in an `examples` folder. I think I would remove the Django File and just have it return the temporary file directly or a string. Dependency injection might be overkill.

msabramo:
I have an idea but I think it can wait until after 0.90. 

Idea is to add option to use and/or fallback to http://docverter.com instead of local pandoc. This could be huge, because then you could possibly use pypandoc without having pandoc installed.

bebraw:
> I like the idea of renaming services.py and putting it in an examples folder. I think I would remove the Django File and just have it return the temporary file directly or a string. Dependency injection might be overkill.

Ok, let's go with that then. I'll do the change for 0.90.

PR #33, opened 2015-02-01
README.rst: Remove "Django" example

msabramo:
I find this example to be confusing, because:

- It's not really Django-specific though it specifically mentions Django. 
- The implementation is in a separate file that is not shown. This is the part that uses pypandoc. The code in README.md uses an API defined in that sample file, which is an API not provided by pandoc. 
- It's too complex for something to show in a README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,21 +40,6 @@ Please refer to `pandoc -h` and the [official documentation](http://johnmacfarla
 
 See also [pyandoc](http://pypi.python.org/pypi/pyandoc/) for an alternative implementation.
 
-## Django Service Example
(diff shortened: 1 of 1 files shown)


bebraw:
I could push the example service below `/examples` and get rid of the Django specific bit through dependency injection.

Do you have anything else in mind before we do a release? See #30.

msabramo:
I like the idea of renaming `services.py` and putting it in an `examples` folder. I think I would remove the Django File and just have it return the temporary file directly or a string. Dependency injection might be overkill.

PR #38, opened 2015-02-06
Allow Markdown extensions in output format

flesser:
From http://johnmacfarlane.net/pandoc/README.html#general-options:
> Markdown syntax extensions can be individually enabled or disabled by appending +EXTENSION or -EXTENSION to the format name. So, for example, ``markdown_strict+footnotes+definition_lists`` is strict markdown with footnotes and definition lists enabled.

Previously, pypandoc did not allow for such an output format to be given.

### DIFF
--- a/pypandoc.py
+++ b/pypandoc.py
@@ -51,9 +51,12 @@ def _convert(reader, processor, source, to,
             'Invalid input format! Expected one of these: ' +
             ', '.join(from_formats))
 
-    if to not in to_formats:
+    # Markdown syntax extensions can be individually enabled or disabled by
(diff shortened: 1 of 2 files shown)


msabramo:
This looks useful!

I wonder if you could add a test - see https://github.com/bebraw/pypandoc/blob/master/tests.py

bebraw:
Yeah, it would be good to have basic tests for this. Please add some and I'll merge your PR.

Thanks for your contribution!

flesser:
Sorry for the late answer. I just added a basic test for it.

bebraw:
@flesser No problem. I merged the feature. You should find it in `0.9.2`. Thanks for your contribution!

PR #41, opened 2015-02-26
Allow Markdown extensions in input format

PhilippWendler:
This was added in cb5a70b, but only for the output format. This adds it also for the input format, where pandoc allows the same way of specifying syntax extensions. A basic test is also included.


bebraw:
Thanks! I published 0.9.3 with your contribution in it.

PhilippWendler:
Wow, that was fast. Thank you.

PR #44, opened 2015-03-09
Support writing output to a file

JanSchulz:
Output writing to a file is needed for some filetypes (e.g. odt, docx).
Closes: #43

Also remove idea/pycharm project files from git

PR #46, opened 2015-03-09
Newer pandoc

### LINKED ISSUE #45: use newer pandoc
See here for the problem: https://github.com/bebraw/pypandoc/pull/44#issuecomment-77880052
-> travis only installs 0.9, but pandoc > 0.10 is needed to support extensions

There are several possible ideas to fix this (apart from "build pandoc from source)
* use IPythons version (see their travis config: https://github.com/ipython/ipython/blob/master/.travis.yml -> this will break if IPYthon removes this version, but only 19MB download
* use the one from RStudio (travis uses that in their r environment): https://github.com/travis-ci/travis-build/pull/386/files#diff-5b7e1aa01f1c6f058df6e0b9e2486c62R370 -> probably stable, but is a 100+MB download
* use a language=r travis config, which has a newer pandoc version and install python via conda :-)

JanSchulz:
This includes #44

Fixes #45

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,7 @@
 dist/
 pypandoc.egg-info/
 .tox/
+
+## IDEs
+# pycharm
+.idea/
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,17 @@
 language: python
 sudo: false
-addons:
-  apt_packages:
-    - pandoc
 env:
-  - TOXENV=py26
-  - TOXENV=py27
-  - TOXENV=py33
-  - TOXENV=py34
-  - TOXENV=pypy
-  - TOXENV=flake8
+  global:
+    - PATH=$TRAVIS_BUILD_DIR/pandoc:$PATH
+  matrix:
+    - TOXENV=py26
+    - TOXENV=py27
+    - TOXENV=py33
+    - TOXENV=py34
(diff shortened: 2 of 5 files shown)


bebraw:
Awesome! Let's go with this.

Thanks a lot @JanSchulz.

JanSchulz:
ok, I missed the flake8 warnings...

Do you really want to have linewrapping at 80? that gets really ugly in tests.py, as the variables are quite long...

bebraw:
> Do you really want to have linewrapping at 80? that gets really ugly in tests.py, as the variables are quite long...

That's true. I did little refactoring there to work around this. I suppose it wouldn't hurt to bump the limit to something like 100-120. Good point there.

bebraw:
@JanSchulz Went with 120. A little saner now. :)

JanSchulz:
@bebraw Thanks for cleaning up after me! Will you upload this as a new version?

bebraw:
@JanSchulz Yeah, done. `0.9.4` contains your work. :+1:

JanSchulz:
@bebraw Thanks!

JanSchulz:
And it is used: https://github.com/JanSchulz/knitpy/blob/master/setup.py :-)

msabramo:
Nice work!

Issue #49, opened 2015-03-18
check returntype of pandoc


JanSchulz:
```
In[7]: output = pypandoc.convert('<h1>Primary Heading</h1>',
    'md', format='html',
    extra_args=['--atx-headers', "--blah"])
pandoc: unrecognized option `--blah'
Try pandoc --help for more information.
In[8]: output
Out[8]: u''
```

bebraw:
Closed by #50.

PR #50, opened 2015-03-18
Check return code and some unicode handling

JanSchulz:
Check the return code of the pandoc call, so that we can raise if there was an error

Closes: #49

Also make sure that we handle unicode in filters-as-a-string and encoded strings and make sure that we always return unicode.


JanSchulz:
finaly... Please tell me if I should split this PR into two or three for the individual issues...

xysmas:
Thanks! I forgot about the 2.7 users when I made the filter handler.

bebraw:
Thanks a lot! 0.9.6 contains the fixes.

Issue #51, opened 2015-04-13
Add a method to get the pandoc version


JanSchulz:
In knitpy, I have the need to ensure that the used pandoc version is higher than 1.10  (and probably 1.12, similar to ipython nbconvert).

ipython has such a version check: https://github.com/jupyter/jupyter_nbconvert/blob/master/jupyter_nbconvert/utils/pandoc.py#L58, maybe that could be copied to pypandoc?

bebraw:
Yeah, feel free to make a PR for that.

JanSchulz:
done in #53

bebraw:
Merged. I published `0.9.7` with the feature in it.

PR #53, opened 2015-04-14
Implement get_pandoc_version

JanSchulz:
Mostly stolen and adapted from https://github.com/jupyter/jupyter_nbconvert/blob/master/jupyter_nbconvert/utils/pandoc.py#L58

Closes: #51


bebraw:
Thanks a lot!

Issue #57, opened 2015-06-09
Converting from .docx / .odt


dlukes:
Pandoc is also able to take e.g. a `.docx` / `.odt` file and output plain text / markdown / html etc., but pypandoc's current approach of first reading the input file in memory as plain text (`pypandoc.py#147`) and then piping the string to pandoc makes it impossible to take advantage of this (obviously since `.docx` / `.odt` is a compressed archive, not a plain text file).

Maybe this is by design (?); if not, it would be great if pypandoc supported this part of pandoc functionality as well :) (TBH, I don't fully understand why it's necessary to read in the input file and pipe it to pandoc instead of just providing it to the pandoc subprocess call as an argument, but I have only taken a glance at the code and there may be ramifications I'm not aware of...)

Great job otherwise!

bebraw:
I wouldn't mind getting this fixed.

Would it be possible for you to prepare a PR to fix the issue? I don't mind if the implementation gets changed (skips reading to memory initially). The code has grown organically and there are knots like that. 

The test suite is somewhat extensive. Just make sure to add `.docx` / `.odt` case there.

Btw if you have any idea on how to work with Travis, please check out #55. I'm a little stumped with that.

dlukes:
OK, I'll see what I can do and get back to you!

I don't know much about Travis though, unfortunately...

bebraw:
> I don't know much about Travis though, unfortunately...

No probs. :+1:

PR #58, opened 2015-06-09
Enable input from non-plain-text files

dlukes:
See issue #57.

This works by letting pandoc itself handle input from any file (instead of first reading the file into a string and then piping it to the pandoc subprocess via STDIN).

Additionally:

- I suggest that temp files in the test suite be wrapped with `with` or `try ... except ... finally`, so that they're destroyed in spite of exceptions (see c95db7d)
- on my machine at least, the test suite fails on both python 2.7 and 3.4 with the following error (irrespective of the modifications in this pull request):

```
ERROR: test_conversion_with_citeproc_filter (__main__.TestPypandoc)
  File "./tests.py", line 133, in test_conversion_with_citeproc_filter
    outputfile=None, filters=filters)
  File "/home/dvl/cnk/src/pypandoc/pypandoc.py", line 94, in convert
  File "/home/dvl/cnk/src/pypandoc/pypandoc.py", line 135, in _convert
  File "/home/dvl/cnk/src/pypandoc/pypandoc.py", line 208, in _process_file
    'Pandoc died with exitcode "%s" during conversation: %s' % (p.returncode, stderr)
RuntimeError: Pandoc died with exitcode "83" during conversation: pandoc: Error running filter pandoc-citeproc
fd:4: hClose: resource vanished (Broken pipe)
```

Do let me know if anything needs to be changed in order to accept the pull request! :)

### DIFF
--- a/pypandoc.py
+++ b/pypandoc.py
@@ -96,7 +96,7 @@ def convert(source, to, format=None, extra_args=(), encoding='utf-8',
 
 def _convert(reader, processor, source, to, format=None, extra_args=(), encoding=None,
              outputfile=None, filters=None):
-    source, format = reader(source, format, encoding=encoding)
+    source, format, input_type = reader(source, format, encoding=encoding)
 
     formats = {
         'dbk': 'docbook',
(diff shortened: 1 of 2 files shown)


bebraw:
> on my machine at least, the test suite fails on both python 2.7 and 3.4 with the following error (irrespective of the modifications in this pull request):

Yeah, it expects citeproc plugin to be installed. No need to worry about that.

I'll have a better look at the morning and get a version out.

Thanks a lot!

dlukes:
> it expects citeproc plugin to be installed

Oh, of course, I should've known :)

bebraw:
I pushed `0.9.8` out there containing your contribution. Thanks a lot!

dlukes:
Thanks! :)

Issue #63, opened 2015-08-09
Please add tests and examples to the released tarball


valholl:
It would be useful to have tests and examples in the released tarball, to be able to sanity check the module before using it (and as documentation): if you are ok with the idea I can prepare a pull request which adds the proper lines to MANIFEST.in.

bebraw:
Sure, sounds nice.

PR #64, opened 2015-08-09
Add tests and examples to distribution.

valholl:
Here it is, as per issue #63. Thanks


bebraw:
Cool, just published a version with the change. :+1:

valholl:
Great, thanks

PR #68, opened 2015-10-04
[WIP] Fix travis

rossant:


JanSchulz:
I'm not sure, but it could be that the reason for that wget is gone. at least the jupyter/nbconvert repo also doesn't use it anymore: https://github.com/jupyter/nbconvert/blob/master/.travis.yml

.travis.yml
```diff
@@ -10,8 +10,11 @@ env:
...
-before_install:
-  - 'wget https://7de4dfdec62155b49b44-d726a73613a1989d29b147f20996e7c1.ssl.cf2.rackcdn.com/pandoc-1.12.3-linux-debian-x86_64.zip && unzip pandoc-1.12.3-linux-debian-x86_64.zip;'
+services:
+  - docker
+addons:
+  apt_packages:
+    - pandoc
 install:
   - travis_retry pip install tox
 script:
```
JanSchulz: I'm not sure if that's automatically done in one of the test scripts, but IMO it would be good to add a `pandoc --version` before the tox call, just to make it reproduceable.

.travis.yml
```diff
@@ -1,5 +1,5 @@
 language: python
-sudo: false
+sudo: required
 env:
   global:
     - PATH=$TRAVIS_BUILD_DIR/pandoc:$PATH
```
JanSchulz: The path doesn't need to be adjusted anymore if the wget line is removed (pandoc was downloaded and extracted to `$TRAVIS_BUILD_DIR/pandoc`...

On the other hand, it could be nice to test against multiple pandoc versions...

JanSchulz:
I'm not sure if that's automatically done in one of the test scripts, but IMO it would be good to add a `pandoc --version` before the tox call, just to make it reproduceable.

.travis.yml
```diff
@@ -10,8 +10,12 @@ env:
     - TOXENV=py34
     - TOXENV=pypy
     - TOXENV=flake8
-before_install:
-  - 'wget https://7de4dfdec62155b49b44-d726a73613a1989d29b147f20996e7c1.ssl.cf2.rackcdn.com/pandoc-1.12.3-linux-debian-x86_64.zip && unzip pandoc-1.12.3-linux-debian-x86_64.zip;'
+services:
+  - docker
```
JanSchulz: According to http://docs.travis-ci.com/user/docker/, this is needed if one wants to use docker calls in the build process, but this travis run doesn't use one. Is there are reason for this? If so it's probably good to add a comment so that others don't stumble about it like I just did :-)

rossant: I *think* the reason for this is to have a newer Ubuntu; not using docker results in travis using 12.04, whereas activating docker enables 14.04 for some reason... But I may be mistaken.

If we use apt we need a recent Ubuntu in order to have a relatively recent version of pandoc.

bebraw:
You might need to install `pandoc-citeproc` as well based on the error.

rossant:
feel free to kill build 105: I think we don't need to install the full LaTeX distribution...

bebraw:
Yeah, did that. :+1:

rossant:
thanks! locally I just have one test failing on Ubuntu:

```

ERROR: test_conversion_from_non_plain_text_file (tests.TestPypandoc)
  File "/home/cyrille/anaconda3/lib/python3.4/site-packages/_pytest/python.py", line 326, in fget
    return self._obj
AttributeError: 'Module' object has no attribute '_obj'

During handling of the above exception, another exception occurred:

  File "/home/cyrille/git/pypandoc/tests.py", line 214, in test_conversion_from_non_plain_text_file
    received = pypandoc.convert(name, to='rst')
  File "/home/cyrille/git/pypandoc/pypandoc.py", line 94, in convert
  File "/home/cyrille/git/pypandoc/pypandoc.py", line 119, in _convert
    ', '.join(from_formats))
RuntimeError: Invalid input format! Expected one of these: docbook, haddock, html, json, latex, markdown, markdown_github, markdown_mmd, markdown_phpextra, markdown_strict, mediawiki, native, opml, rst, textile
```

bebraw:
That

> AttributeError: 'Module' object has no attribute '_obj'

is a strange one. Any idea what might cause it? Given it's on 3.4, maybe that's on our side.

## tok Error

> for tok in out_lines[0].split():

implies that `pandoc -v` doesn't give any output at all. We could at least raise an error if `out_lines` doesn't have any content.

Maybe the difference is that we're doing `pandoc -v` instead of `pandoc --version`? I suspect `pandoc -v` is a newer feature. This could explain the failure in this case.

rossant:
There's a problem with HOME and pandoc-citeproc: https://github.com/jgm/pandoc-citeproc/issues/35

bebraw:
Yeah, by the looks of it pandoc-citeproc doesn't give us much to work with... I know it's not ideal but for now we could detect if we're running on Travis and skip citeproc specific checks. Something is better than nothing after all.

bebraw:
Can you do

> ['pandoc', '--version']

instead of

> ['pandoc', '-v']

?

I have a feeling that might fix the last error.

bebraw:
Based on your output even `pandoc -v` should work. Why `for tok in out_lines[0].split():` fails in both cases?

That `test_conversion_from_non_plain_text_file` failure is probably explained by old pandoc. We're converting to docx after all.

rossant:
OK I'm out of luck... feel free to merge (or not), I think this is *slightly* better than before...

bebraw:
Yeah. Thanks for the effort, I'll merge this.

Perhaps there's some way to get those failing tests disabled on Travis. :+1:

PR #71, opened 2015-10-14
don't require pandoc to install pypandoc

### LINKED ISSUE #70: pypandoc install fails of pandoc isn't installed
This is pretty easy to repro.  Just do the following in a virtualenv on a machine where pandoc is not installed:

```
pip install pypandoc
```

I think errors related to pandoc not being installed should be deferred until execution time and should not prevent individuals from installing the library.  The particular use case where this is really impacting me currently is when trying to use setuptools-markdown (https://github.com/msabramo/setuptools-markdown).  Here, it is reasonable for a maintainer to install pandoc to convert markdown to restructured text when releasing to pypi.  Unfortunately, right now we need to throw the baby out with the bathwater because `setup_requires` also installs dependencies for every client installing the package.

Since most clients don't have pandoc installed (and they shouldn't need it), using `setuptools-markdown` isn't an option and we can't have nice things.  The stack trace that shows up seems to indicate that fixing this might be as easy as putting a try clause around your setup.py:

```
Collecting pypandoc
  Downloading pypandoc-1.0.4.tar.gz
    ---------------------------------------------------------------
    An error occurred while trying to run `pandoc`
    Maybe try:
        sudo yum install pandoc
    See http://johnmacfarlane.net/pandoc/installing.html
    for installation options
    ---------------------------------------------------------------
      File "<string>", line 20, in <module>
      File "/tmp/pip-build-SRTyxW/pypandoc/setup.py", line 6, in <module>
        long_description = pypandoc.convert('README.md', 'rst')
      File "pypandoc.py", line 94, in convert
      File "pypandoc.py", line 114, in _convert
        from_formats, to_formats = get_pandoc_formats()
      File "pypandoc.py", line 276, in get_pandoc_formats
        raise OSError("You probably do not have pandoc installed.")
    OSError: You probably do not have pandoc installed.
    Complete output from command python setup.py egg_info:
    ---------------------------------------------------------------
    
    An error occurred while trying to run `pandoc`
    
    Maybe try:
    
    
    
        sudo yum install pandoc
    
    See http://johnmacfarlane.net/pandoc/installing.html
    
    for installation options
    
    ---------------------------------------------------------------
    
    
    
      File "<string>", line 20, in <module>
    
      File "/tmp/pip-build-SRTyxW/pypandoc/setup.py", line 6, in <module>
    
        long_description = pypandoc.convert('README.md', 'rst')
    
      File "pypandoc.py", line 94, in convert
    
      File "pypandoc.py", line 114, in _convert
    
        from_formats, to_formats = get_pandoc_formats()
    
      File "pypandoc.py", line 276, in get_pandoc_formats
    
        raise OSError("You probably do not have pandoc installed.")
    
    OSError: You probably do not have pandoc installed.
    
    ----------------------------------------
    Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-build-SRTyxW/pypandoc
```

posborne:
This changes closes #70 by allowing for pypandoc to install
without pandoc being present on the system.  This allows for
the library to be installed but for clients to handle the
fact that pandoc is not installed in their applications at
runtime.

If pandoc is not present during installation, the warning
from pypandoc is still printed informing users how pandoc
can be installed.

Signed-off-by: Paul Osborne <osbpau@gmail.com>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,11 @@
 import pypandoc
 from setuptools import setup
 
-long_description = pypandoc.convert('README.md', 'rst')
+try:
+    long_description = pypandoc.convert('README.md', 'rst')
+except OSError:
+    # pandoc is not installed, fallback to using raw contents
+    long_description = open('README.md').read()
 
 module = pypandoc
 setup(


bebraw:
@posborne I published `1.0.5` with the change. Hopefully pypandoc works for you now. :+1:

Issue #16, opened 2014-12-23
Idea: Create wheels bundled with pandoc


msabramo:
I think the one tricky thing about this module is that it relies on having pandoc installed and that's not Python. 

I wonder if you can create wheels on OS X and Linux that bundle a pre-built pandoc binary. You'd have to check with John MacFarlane to see if he's ok with this, but I think it will meet the license (GPL) as long as you include a notice with a link to the source code. 

I mention OS X and Windows and not Linux because PyPI won't let you upload binary wheels for Linux, because it's very hard to build one that works reliably across different Linux versions.

bebraw:
What do you want to do with this issue?

msabramo:
@bebraw: Take a look at https://github.com/bebraw/pypandoc/pull/37 -- hopefully you have a Mac to test it on.

Not sure how you do the same on Linux, but I suspect that it's possible.

JanSchulz:
rstudio also bundles pandoc

JanSchulz:
the used builds for rstudio are here: https://s3.amazonaws.com/rstudio-buildtools/pandoc-1.13.1.zip (and probably other versions, as they are used in in travis builds: https://github.com/travis-ci/travis-build/pull/386/files#diff-5b7e1aa01f1c6f058df6e0b9e2486c62R370). They include linux, mac and windows versions...

Not sure if that is "nice", but:

```
if build_type == wheel:
   wget RStudio_pandoc.zip
   untip the right one into a special dir
   include the files in the wheel build
```

And in the code then add something, which takes into account the filepath to that dir...

pombreda:
:+1:  for vendoring pandoc binaries. I actually have code for a soon to be released library that does just that.

JanSchulz:
So, should this happen on all wheel builds? I'm willing to give this a try... It seems that it needs a special treatment for all supported archs as rstudio doesn't ship newer versions than 1.13.

On windows:
* create `%temp_dir%` + `%temp_dir%\pandoc`
* use request or whatever to downlaod the latest pandoc version from pandocs download page to %temp_dir%
* `msiexec /a %temp_dir%\pandoc-1.15.0.6-windows.msi /qb TARGETDIR=%temp_dir%\pandoc
* copy pandoc and pandoc-citeproc to a the right dir in the source (where is that? in other repos I would add `pypandoc/{__init__.py,bin/} and copy to bin, but pypadoc is structured without a package dir)
* include it in the wheel build...
[...]

Issue #72, opened 2015-10-19
Specifying path to pandoc


wcaleb:
Is there some way to specify the path to pandoc within a script? I am importing pypandoc into a Flask app that runs on an Apache server, and since the server runs as a special Apache user, the app can't seem to find pandoc even though it's installed in my own user path. I know [pyandoc](https://github.com/kennethreitz/pyandoc) had a way around this with a variable for the pandoc path, but can't tell if there's a way to manage that here.

JanSchulz:
As far as I know there isn't. Pandoc is expected to be on the PATH, so if you have any way to change that for your script it should be fine...

Having a way to specify the pandoc path would be a nice first strep to include the pandoc binary in a wheel...

bebraw:
PR is welcome. :+1:

JanSchulz:
See #73 for a PR implementing this

wcaleb:
Thanks for working on this!

bebraw:
@wcaleb Just to clarify, it should work based on `PYPANDOC_PANDOC` now. Set that through env.

gwgundersen:
I have an issue that is, I think, related to this. I have pandoc and pypandoc both installed, and everything works from the REPL. But when I try to execute pypandoc from my Flask application, I get a "No pandoc was found" error. How is `PYPANDOC_PANDOC` related to fixing this issue? Thanks.

gwgundersen:
For anyone else who finds this question searching for a solution, just set the environment variable at the Flask instance startup, e.g.:

```
# __init__.py

import os
os.environ.setdefault('PYPANDOC_PANDOC', '/usr/local/bin/pandoc')
```

bebraw:
@gwgundersen Maybe that's worth adding to the README?

gwgundersen:
Okay, see PR #99 for my attempt at explaining this.

PR #37, opened 2015-02-04
Include and use pandoc in wheel for OS X

msabramo:
I copied a Homebrew-built pandoc and libgmp, using delocate to rewrite
the pandoc binary to look for the library in ../lib instead of
/usr/local/lib.

This pandoc binary and libgmp dylib get included in wheels.

If we can't find `pandoc` and `sys.platform == darwin` then we use the
darwin pandoc in the wheel.

```
$ brew uninstall pandoc libgmp
...
$ pip uninstall pypandoc; pip install ~/dev/git-repos/pypandoc/dist/pypandoc-0.9.0-py2-none-any.whl
Uninstalling pypandoc-0.9.0:
  Successfully uninstalled pypandoc-0.9.0
Processing ./dev/git-repos/pypandoc/dist/pypandoc-0.9.0-py2-none-any.whl
Requirement already satisfied (use --upgrade to upgrade): setuptools in ./python/virtualenvs/pypandoc-standalone/lib/python2.7/site-packages (from pypandoc==0.9.0)
Installing collected packages: pypandoc

Successfully installed pypandoc-0.9.0

$ ipython
Python 2.7.9 (v2.7.9:648dcafa7e5f, Dec 10 2014, 10:10:46)
Type "copyright", "credits" or "license" for more information.

IPython 2.0.0 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.

In [1]: import pypandoc

In [2]: pypandoc.convert('[google](http://www.google.com/)', 'rst', format='md')
Out[2]: u'`google <http://www.google.com/>`__\n'
```

PR #73, opened 2015-10-20
Include pandoc binaries

JanSchulz:
This is a first version how to include the pandoc binaries. Only windows is currently implemented...

```
python setup.py download_pandoc
python setup.py bdist_wheel
```

Will first download the pandoc binaries and put them in `pypandoc/files`. The bdist_wheel step then includes the files in the wheel package. Pandoc now searches for pandoc first on the path and then on the new location in the package and takes the bigger pandoc version (equal will use the one on path). 

I move the `pypandoc.py` file to `pypandoc.__init__.py` as I wanted to have all files in a subdir and not directly in site_packages.

replaces: #37 
Closes: #16 
Closes:: #72: env var for path to pandoc


JanSchulz:
If this is taken (aka the move from pypandoc.py to pypandoc/__init__.py, then I think it makes sense to split out the py3compat stuff to a different file...

bebraw:
> If this is taken (aka the move from pypandoc.py to pypandoc/init.py, then I think it makes sense to split out the py3compat stuff to a different file...

I'm not exactly sure how people like to deal with this in Python but that sounds reasonable.

JanSchulz:
Yay, it's green...

JanSchulz:
Ok, this has one big drawback right now: it builds a wheel with 'any' in the filename, so no windows/darwin specific wheel... Lets see...

This includes now the split of py3compat stuff into a new file. If someone with a macosx system wants to try the macosx part or somone the linux part, I'd love to include that...

JanSchulz:
Ok, now we get platform specific wheels, but the egg seems to be still platform agnostic. Not sure what to do with it, I've never used eggs...

JanSchulz:
Fixed the version number screwup and also fixed some commits which mixed wrong stuff...

bebraw:
Do you expect that the binary distributions will be available through PyPI? I don't mind merging your work. Just curious.

JanSchulz:
I would be very happy if there would be binary wheels on PyPi.

I see the problem that pandoc is GPL, so that means that I should add the license to some place of the package and theoretically someone could then ask you to send a physical cd with the sources...

http://stackoverflow.com/questions/5437501/can-i-use-gpl-software-binaries-in-commercial-environment

setup.py
```diff
@@ -9,6 +22,111 @@
     # pandoc is not installed, fallback to using raw contents
     long_description = open('README.md').read()
 
+# Uses sys.platform keys, but removes the 2 from linux2
+# Adding a new plattform means implemnting unpacking in "DownloadPandocCommand" and adding the URL here
```
bebraw: typo: implemnting -> implementing

setup.py
```diff
@@ -9,6 +22,111 @@
...
+    "win32": "https://github.com/jgm/pandoc/releases/download/1.15.1.1/pandoc-1.15.1.1-windows.msi",
+    #"linux": "https://github.com/jgm/pandoc/releases/download/1.15.1/pandoc-1.15.1-1-amd64.deb",
+    #"darwin": "https://github.com/jgm/pandoc/releases/download/1.15.1/pandoc-1.15.1-osx.pkg"
+}
+
+class DownloadPandocCommand(Command):
+
+    """Download pandoc"""
+
+    description = "downloads a pandc release and adds it to the package"
```
bebraw: pandc -> pandoc

setup.py
```diff
@@ -9,6 +22,111 @@
...
+    def finalize_options(self):
+         pass
+
+    def _unpack_win32(self, filename):
+        targetfolder = os.path.join(os.path.dirname(os.path.realpath(__file__)), "pypandoc", "files")
+        print("* Unpacking %s to tempfolder..." % (filename))
+
+        tempfolder = tempfile.mkdtemp()
+
+        # Make sure the targetfolder exists...
```
bebraw: targetfolder -> target folder perhaps

bebraw:
> I would be very happy if there would be binary wheels on PyPi.

How do other packages deal with binaries? What's the standard?

JanSchulz:
> How do other packages deal with binaries? What's the standard?

Actually I haven't seen any packages which do/need this.

IMO this is the right approach, at least when staying in the python world (e.g. distributor should not use included binaries and maybe conda should work by building a pandoc package which puts pandoc in PATH (script dir?). I'm not really happy with the wheel hack, but I currently see no alternatives (hm, maybe monkeypatching `self.distribution.has_ext_...()` whatever distributin is in that context).

bebraw:
Do you know how big the binaries are? Do you need some specific environment to upload them? I'm on OS X myself btw. I can likely test that Darwin part if needed.

JanSchulz:
The two static files on windows (`pandoc.exe` and `pandoc_citeproc.exe`) are ~80MB... The `pypandoc-1.0.3-cp27-none-win_amd64.whl` is ~20MB

The package only includes the prebuilt native binaries and otherwise only python code, so no special built env is needed, but you do currently need a OS for each supported plattform (so windows, macosx and it's probably ok to assume that linux users can take care for themself?)

Upload (using [twine](https://pypi.python.org/pypi/twine/)) is then 
```
twine upload dist/*
```

I can build windows binaries and send them to you via dropbox...

There is also https://github.com/conda-forge/conda-smithy which can hopefully gain some other PyPi uploads, too...

A completely different way would be to build the packages by adding a target platform (`python setup.py download_pandoc windows`) and then renaming the built wheel. That should work, as far as I know the only target specific thing is the name of the wheel.

So a `release.sh`:

```
python setup.py download_pandoc windows
python setup.py bdist_wheel
# rename wheel
# remove prebuilt binaries
python setup.py download_pandoc linux
...
```

JanSchulz:
Ok, I managed to setup a smithy for pypandoc (but I will expand that to other packages) which uploads conda packages to my channel (I need pypandoc for my knitpy project).

https://github.com/JanSchulz/package-builder/

I think adding uploads to pypi will be possible by extending the [build scripts](https://github.com/JanSchulz/package-builder/tree/master/recipes/pypandoc) with some additional commands:

* download the binaries
* build the wheel
* upload to PyPi
* cleanup
* install for conda building...

I currently don't need it for knitpy (I will only build one wheel, which I will let the normal travis build upload), but I can setup the thing for pypi and you can fork and change the tokens for your accounts?

bebraw:
@JanSchulz What if I upgraded you into a maintainer status? I'm absolutely certain you would be great at that role given your contributions so far. That way you could publish versions whenever you need and so on. I can of course remain in an advisory role but given I don't do any Python development beyond this project, I'm a little out of the loop so to speak.

JanSchulz:
@bebraw I would be happy to be added as a maintainer! :-) 

I have to say that I'm happy with pypandoc as it is (minus the not included binaries): it does exactly what I need and I currently wouldn't do much development on pypandoc if there is no pressing issue/request coming up.

bebraw:
@JanSchulz Alright. I added you to the GitHub project collaborators so you should have access to the repository. You are also an owner of the PyPI package now so you can release as you wish.

I'm not expecting any wonders. Just responding to the issues every once in a while will do just fine. If you have time/interest at some point, there are a couple of older issues to study.

JanSchulz:
Thank you! I will try to setup the wheel building. 

If you want to merge this before and make a release, please go ahead.

bebraw:
@JanSchulz I published `1.1.0` to signify the start of the new series. There's just a source dist for now. Feel free to proceed with the binaries as you want.

JanSchulz:
@bebraw Could you add a `_unpack_darwin` function to setup.py? That's what currently missing to push it...

http://stackoverflow.com/questions/11298855/how-to-unpack-and-pack-pkg-file

Should be mostly c&p from here or if that's possible to directly unpack into the right dir...
```
    def _unpack_win32(self, filename):
        targetfolder = os.path.join(os.path.dirname(os.path.realpath(__file__)), "pypandoc", "files")
        print("* Unpacking %s to tempfolder..." % (filename))

        tempfolder = tempfile.mkdtemp()

        # Make sure targetfolder exists...
        try:
            os.makedirs(targetfolder)
        except OSError:
            pass # dir already exists...

        cmd = ["msiexec", "/a", filename, "/qb", "TARGETDIR=%s" % (tempfolder)]
        # if only 3.5 is supported, should be `run(..., check=True)`
        subprocess.check_call(cmd)

        # pandoc.exe, pandoc-citeproc.exe, and the COPYRIGHT are in the Pandoc subfolder
        for exe in ["pandoc.exe", "pandoc-citeproc.exe", "COPYRIGHT.txt"]:
            src = os.path.join(tempfolder, "Pandoc", exe)
            dst = os.path.join(targetfolder, exe)
            print("* Copying %s to %s ..." % (exe, targetfolder))
            print("%s, %s" % (src, dst))
            shutil.copyfile(src, dst)

        # remove temporary dir
        shutil.rmtree(tempfolder)
        print("* Done.")
```

bebraw:
@JanSchulz I sketched out something. I ran into a problem, though:

```bash
pypandoc * [master] $ python setup.py download_pandoc
running download_pandoc
* Downloading pandoc from https://github.com/jgm/pandoc/releases/download/1.15.1/pandoc-1.15.1-osx.pkg ...
  File "setup.py", line 184, in <module>
    cmdclass=cmd_classes
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/distutils/core.py", line 151, in setup
    dist.run_commands()
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/distutils/dist.py", line 953, in run_commands
    self.run_command(cmd)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/distutils/dist.py", line 972, in run_command
    cmd_obj.run()
  File "setup.py", line 126, in run
    with urlopen(url) as response:
AttributeError: addinfourl instance has no attribute '__exit__'
```

Any idea what might cause that?

JanSchulz:
Seems that urlopen on py27 isn't implemented as a context manager :-(

http://stackoverflow.com/questions/30627937/tracebaclk-attributeerroraddinfourl-instance-has-no-attribute-exit


```
        with urlopen(url) as response:
                with open(filename, 'wb') as out_file:
                    shutil.copyfileobj(response, out_file)
```

-> 

```
        response = urlopen(url)
        with open(filename, 'wb') as out_file:
            shutil.copyfileobj(response, out_file)
```

JanSchulz:
If you are at it: in setup.py the checked exception needs to change from `OSError` to  `RuntimeError`...
-> https://travis-ci.org/JanSchulz/package-builder/builds/86807278

Damn :-(

bebraw:
> If you are at it: in setup.py the checked exception needs to change from OSError to RuntimeError...
-> https://travis-ci.org/JanSchulz/package-builder/builds/86807278

Done.

Getting close with Darwin.

JanSchulz:
Wow, that was fast! :-)

this line can probably also go, it's a leftover from my debugging session: `print("%s, %s" % (src, dst))`

JanSchulz:
Ok, the assert for 64 bit can also go, at least on windows:

```
λ file ..\pypandoc\pypandoc\files\pandoc.exe
..\pypandoc\pypandoc\files\pandoc.exe: PE32 executable for MS Windows (console) Intel 80386 32-bit
```

-> So it will run fine on 32 bit. I suspect that's the same for OSX...

bebraw:
Ok, I did those tweaks. `python setup.py download_pandoc` appears to work for me now.

JanSchulz:
Yay, and there come the wheels:
https://testpypi.python.org/pypi/pypandoc (right now only windows yet...)
https://anaconda.org/janschulz/pypandoc/files (deleted most of them again, waiting for the macosx build to succeed...)

bebraw:
Ok, I have no idea how Linux would work out but if someone needs that, I guess that person will implement the portion.

JanSchulz:
Linux would be nice, as that would give someone using our wheels/conda packages the ability to use them on travis without working around older pandoc versions...

It's actually this two commands:
```
ar x xxx.deb
tar xzf data.tar.gz
```

(and to figure out how to make `ar` do it in a temp directory...)

PR #77, opened 2015-10-23
Expose pandoc path

JanSchulz:
This switches back to use `OSError` if pandoc is not found (was briefly `RuntimeError`, which could be considered a API breakage :-() and exposes the computed pandoc path as an API call.

### DIFF
--- a/pypandoc/__init__.py
+++ b/pypandoc/__init__.py
@@ -12,7 +12,7 @@ from .py3compat import string_types, cast_bytes, cast_unicode
 __author__ = u'Juho Vepsäläinen'
 __version__ = '1.1.0'
 __license__ = 'MIT'
-__all__ = ['convert', 'get_pandoc_formats', 'get_pandoc_version']
+__all__ = ['convert', 'get_pandoc_formats', 'get_pandoc_version', 'get_pandoc_path']
 
 
 def convert(source, to, format=None, extra_args=(), encoding='utf-8',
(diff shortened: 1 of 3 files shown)


bebraw:
Feel free to cut a release. :+1:

JanSchulz:
Done, currently running the wheel / conda builds :-)

JanSchulz:
And still the wrong long description...? Hm, is twine the problem?

bebraw:
I am not sure. I patched the long description by hand for now.

Even if it fails to process it through pandoc, it still should use the contents of the entire file for `long_description`. Could this be something Python 3 related? I don't know PyPI well enough to tell.

JanSchulz:
I think it's twine: it uses `pkginfo.SDist` and that lists only `pypandoc` in the `description`... Not sure why :-(

bebraw:
Ok. Maybe you can poke some Twine guys about it.

JanSchulz:
Ok, I think I found the problem: pypandoc returned `\r\n` lineendings, which somehow got transfered into two lineendings, resulting in a blank line, which meant that the first headline ("pypandoc") was separated from the rst underline by a blank line, which sdist translates into the end of the description...

bebraw:
That sounds strange. I guess one way to solve that would be to convert back to something normal. Of course it's a good question why it outputs those line endings in the first place. Is there an encoding option or such we could use?

JanSchulz:
I think it's just pandoc but the strange thing is why setuptools doubles the linefeeds...

bebraw:
Ok. Can we monkeypatch it for now?

JanSchulz:
I added a simple replace("\r",""), works Luke a charm :-)

PR #82, opened 2015-11-10
[DO NOT MERGE] Travis work

JanSchulz:
Add python 3.5 to the test matrix. We could probably remove 3.4 or so if that's too much testing, as 3.3 and 3.5 should cover the middle...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,10 +19,7 @@ addons:
       - lmodern
 install:
   - travis_retry pip install tox
-  # download a newer pandoc version...
-  - python setup.py download_pandoc
-
+  
 script:
-  - pandoc -v
-  - python -c "import pypandoc; print(pypandoc.get_pandoc_path(), pypandoc.get_pandoc_version())"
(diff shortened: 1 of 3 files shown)


JanSchulz:
Ok, 3.5 is not yet available: https://github.com/travis-ci/travis-ci/issues/4794

bebraw:
Alright. Feel free to merge once Travis supports it. No objections. :+1:

JanSchulz:
Ok, I give up for now to use our own downloader... I think we should refactor the code, so that the downloader becomes part of the library (e.g `pypandoc.download_pandoc(to_path='.', url=None)` -> you can download to some specific location (`setup.py download_pandoc` would set that to `pypandoc/files`) and also update pandoc if a new release comes out) and the setup just calls it when building wheels. This would make it possible to download pandoc by the user if there was no wheel available.

JanSchulz:
This is now solved in #92, I just committed the py35 test addition there as well.

Issue #91, opened 2016-01-15
Update pandoc


JanSchulz:
https://github.com/jgm/pandoc/releases -> pandoc 1.16.0.2 is current, we ship 1.15... Should probably be done after https://github.com/bebraw/pypandoc/issues/83

bebraw:
Ok, feel free to proceed with it as you see fit. :+1:

Issue #90, opened 2016-01-15
Build on py3.5 fails


JanSchulz:
```
$ obvci_conda_build_dir ./recipes janschulz --channel main --build-condition "python >=3.5,<3.6"
Fetching package metadata: ......
Resolving distributions from 5 recipes... 
Computed that there are 4 distributions from the 5 recipes:
Resolved dependencies, will be built in the following order: 
	matplotlib-1.5.0+532.g3d942f7.dirty-py35_0 (will be built: False)
	pypandoc-1.1.3-py35_0 (will be built: True)
	statsmodels-0.6.1-py35_0 (will be built: True)
	ipyext-0.1.0-py35_0 (will be built: False)
Nothing to be done for matplotlib - it is already on main.
Building  pypandoc-1.1.3-py35_0
Removing old build environment
Removing old work directory
BUILD START: pypandoc-1.1.3-py35_0
Fetching package metadata: ........
Solving package specifications: ..........
The following packages will be downloaded:
    package                    |            build
    ---------------------------|-----------------
    python-3.5.1               |                0        12.7 MB  defaults
    setuptools-19.2            |           py35_0         349 KB  defaults
    wheel-0.26.0               |           py35_1          77 KB  defaults
    pip-7.1.2                  |           py35_0         1.4 MB  defaults
    ------------------------------------------------------------
                                           Total:        14.6 MB
The following NEW packages will be INSTALLED:
    openssl:    1.0.2e-0      defaults
    pip:        7.1.2-py35_0  defaults
    python:     3.5.1-0       defaults
    readline:   6.2-2         defaults
    setuptools: 19.2-py35_0   defaults
    sqlite:     3.9.2-0       defaults
    tk:         8.5.18-0      defaults
    wheel:      0.26.0-py35_1 defaults
    xz:         5.0.5-0       defaults
    zlib:       1.2.8-0       defaults
Removing old work directory
checkout: u'v1.1.3'
==> git log -n1 <==
commit 7f03f316f2015fd8254fb2387681008f2b597444
Author: Jan Schulz <jasc@gmx.net>
Date:   Fri Jan 15 21:27:17 2016 +0100
    pypandoc v1.1.3
==> git describe --tags --dirty <==
v1.1.3
==> git status <==
HEAD detached at v1.1.3
nothing to commit, working directory clean
Package: pypandoc-1.1.3-py35_0
source tree in: /Users/travis/miniconda/conda-bld/work
[...]

Issue #84, opened 2015-11-21
Workaround to make the wheel platform specific


JanSchulz:
See here: http://lucumr.pocoo.org/2014/1/27/python-on-wheels/

```
import os
from setuptools import setup
from setuptools.dist import Distribution

class BinaryDistribution(Distribution):
    def is_pure(self):
        return False

setup(
    ...,
    include_package_data=True,
    distclass=BinaryDistribution,
)
```

bebraw:
If you think it's a worthwhile change, please go ahead. :+1:

Issue #83, opened 2015-11-10
Make pandoc download a regular library function


JanSchulz:
I think we should refactor the code, so that the downloader becomes part of the library (e.g `pypandoc.download_pandoc(to_path='.', url=None)` -> you can download to some specific location (`setup.py download_pandoc` would set that to `pypandoc/files`) and also update pandoc if a new release comes out) and the setup just calls it when building wheels. This would make it possible to download pandoc by the user if there was no wheel available.

PR #92, opened 2016-02-14
Overhaul pandoc handling

JanSchulz:
This does two things:
* It refactors the old setup.py based download function so that it is a user function which can be called by the user and by setup.py
* the new functions gets some default target folder for the download and these folder + a few default platform specific folders get added to the search path

The result should be that if you install the pandoc on your system, pypandoc should find it... And if you don't have a pandoc installed, it is only a `pypandoc.download_pandoc()` away...

If there are a typical install location missing, please leave a comment...

Closes: #91 -> new pandoc version
Closes: #84 -> new cleaner workaround for making a wheel platform specific
Closes: #90 -> new workaround should compile on py35
Closes: #83 -> now available as `pypandoc.download_pandoc([url, targetfolder])`

### DIFF
--- a/pypandoc/pandoc_download.py
+++ b/pypandoc/pandoc_download.py
@@ -22,7 +22,8 @@ PANDOC_URLS = {
     "linux": "https://github.com/jgm/pandoc/releases/download/1.16.0.2/pandoc-1.16.0.2-1-amd64.deb",
     "darwin": "https://github.com/jgm/pandoc/releases/download/1.16.0.2/pandoc-1.16.0.2-osx.pkg"
 }
-INCLUDED_PANDOC_VERSION="1.16.0.2"
+
+INCLUDED_PANDOC_VERSION = "1.16.0.2"
 
 DEFAULT_TARGET_FOLDER = {
     "win32": "~\\AppData\\Local\\Pandoc",
@@ -34,7 +35,9 @@ DEFAULT_TARGET_FOLDER = {
 def _make_executable(path):
     mode = os.stat(path).st_mode
     mode |= (mode & 0o444) >> 2    # copy R bits to X
+    print("* Making %s executeable..." % (path))
     os.chmod(path, mode)
+
 
 def _handle_linux(filename, targetfolder):
 
--- a/pypandoc/pandoc_download.py
+++ b/pypandoc/pandoc_download.py
@@ -18,9 +18,9 @@ except ImportError:
 # Adding a new platform means implementing unpacking in "DownloadPandocCommand"
 # and adding the URL here
 PANDOC_URLS = {
-    "win32": "https://github.com/jgm/pandoc/releases/download/1.15.1.1/pandoc-1.15.1.1-windows.msi",
-    "linux": "https://github.com/jgm/pandoc/releases/download/1.15.1/pandoc-1.15.1-1-amd64.deb",
-    "darwin": "https://github.com/jgm/pandoc/releases/download/1.15.1/pandoc-1.15.1-osx.pkg"
+    "win32": "https://github.com/jgm/pandoc/releases/download/1.16.0.2/pandoc-1.16.0.2-windows.msi",
+    "linux": "https://github.com/jgm/pandoc/releases/download/1.16.0.2/pandoc-1.16.0.2-1-amd64.deb",
+    "darwin": "https://github.com/jgm/pandoc/releases/download/1.16.0.2/pandoc-1.16.0.2-osx.pkg"
 }
 
 
--- a/pypandoc/pandoc_download.py
+++ b/pypandoc/pandoc_download.py
@@ -23,6 +23,12 @@ PANDOC_URLS = {
     "darwin": "https://github.com/jgm/pandoc/releases/download/1.16.0.2/pandoc-1.16.0.2-osx.pkg"
 }
 
+DEFAULT_TARGET_FOLDER = {
+    "win32": "~\\AppData\\Local\\Pandoc",
+    "linux": "~/bin",
+    "darwin": "~/Applications/pandoc"
+}
+
 
 def _handle_linux(filename, targetfolder):
 
@@ -45,7 +51,7 @@ def _handle_linux(filename, targetfolder):
             print("* Copying %s to %s ..." % (exe, targetfolder))
             shutil.copyfile(src, dst)
         src = os.path.join(tempfolder, "usr", "share", "doc", "pandoc", "copyright")
-        dst = os.path.join(targetfolder, "copyright")
+        dst = os.path.join(targetfolder, "copyright.pandoc")
         print("* Copying copyright to %s ..." % (targetfolder))
         shutil.copyfile(src, dst)
     finally:
@@ -102,17 +108,33 @@ def _handle_win32(filename, targetfolder):
 
 
 def download_pandoc(url=None, targetfolder=None):
+    """Download and unpack pandoc
+
+    Downloads prebuild binaries for pandoc from `url` and unpacks it into
+    `targetfolder`.
+
+    :param str url: URL for the to be downloaded pandoc binary distribution for
+        the platform under which this python runs. If no `url` is give, uses
+        the latest available release at the time pypandoc was released.
+
+    :param str targetfolder: directory, where the binaries should be installed
+        to. If no `targetfolder` is give, uses a platform specific user
+        location: `~/bin` on Linux, `~/Applications/pandoc` on Mac OS X, and
+        `~\\AppData\\Local\\Pandoc` on Windows.
+    """
     pf = sys.platform
+
     # compatibility with py3
     if pf.startswith("linux"):
         pf = "linux"
-        assert platform.architecture()[0] == "64bit", "Linux pandoc is only compiled for 64bit."
+        if platform.architecture()[0] != "64bit":
+            raise RuntimeError("Linux pandoc is only compiled for 64bit.")
+
+    if pf not in PANDOC_URLS:
+        raise RuntimeError("Can't handle your platform (only Linux, Mac OS X, Windows).")
 
     if url is None:
-        try:
-            url = PANDOC_URLS[pf]
-        except:
-            raise Exception("No prebuilt pandoc available or not yet implemented for your platform")
+        url = PANDOC_URLS[pf]
 
     filename = url.split("/")[-1]
     if os.path.isfile(filename):
@@ -123,8 +145,10 @@ def download_pandoc(url=None, targetfolder=None):
         response = urlopen(url)
         with open(filename, 'wb') as out_file:
             shutil.copyfileobj(response, out_file)
+
     if targetfolder is None:
-        targetfolder = os.path.join(os.path.dirname(os.path.realpath(__file__)), "pypandoc", "files")
+        targetfolder = DEFAULT_TARGET_FOLDER[pf]
+    targetfolder = os.path.expanduser(targetfolder)
 
     # Make sure target folder exists...
     try:
--- a/pypandoc/pandoc_download.py
+++ b/pypandoc/pandoc_download.py
@@ -22,6 +22,7 @@ PANDOC_URLS = {
     "linux": "https://github.com/jgm/pandoc/releases/download/1.16.0.2/pandoc-1.16.0.2-1-amd64.deb",
     "darwin": "https://github.com/jgm/pandoc/releases/download/1.16.0.2/pandoc-1.16.0.2-osx.pkg"
 }
+INCLUDED_PANDOC_VERSION="1.16.0.2"
 
 DEFAULT_TARGET_FOLDER = {
     "win32": "~\\AppData\\Local\\Pandoc",
@@ -29,6 +30,11 @@ DEFAULT_TARGET_FOLDER = {
     "darwin": "~/Applications/pandoc"
 }
 
+
+def _make_executable(path):
+    mode = os.stat(path).st_mode
+    mode |= (mode & 0o444) >> 2    # copy R bits to X
+    os.chmod(path, mode)
 
 def _handle_linux(filename, targetfolder):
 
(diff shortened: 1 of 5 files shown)


JanSchulz:
@bebraw could you run a few tests on osx? Someone here on linux?

* Does wheel building work? `python setup.py download_pandoc dist_wheel`
* Does `pypandoc.download_pandoc()` work?
* Does it finds the installed pandoc?

Are the default install location ok for your platform: https://github.com/bebraw/pypandoc/pull/92/files#diff-5d5106ab2bc99eacc6d82b01f674f9a3R26

bebraw:
Sure. I'll have a good look around Tuesday. Busy day tomorrow. :+1:

bebraw:
Finally got time to test this:

```bash
$ python setup.py dist_wheel
usage: setup.py [global_opts] cmd1 [cmd1_opts] [cmd2 [cmd2_opts] ...]
   or: setup.py --help [cmd1 cmd2 ...]
   or: setup.py --help-commands
   or: setup.py cmd --help

error: invalid command 'dist_wheel'
```

Am I missing something? I noticed `bdist_wheel` but that didn't help either.

JanSchulz:
IMO there is only `python setup.py bdist_wheel`.

pypandoc/pandoc_download.py
```diff
@@ -0,0 +1,162 @@
...
+PANDOC_URLS = {
+    "win32": "https://github.com/jgm/pandoc/releases/download/1.16.0.2/pandoc-1.16.0.2-windows.msi",
+    "linux": "https://github.com/jgm/pandoc/releases/download/1.16.0.2/pandoc-1.16.0.2-1-amd64.deb",
+    "darwin": "https://github.com/jgm/pandoc/releases/download/1.16.0.2/pandoc-1.16.0.2-osx.pkg"
+}
+
+DEFAULT_TARGET_FOLDER = {
+    "win32": "~\\AppData\\Local\\Pandoc",
+    "linux": "~/bin",
+    "darwin": "~/Applications/pandoc"
```
JanSchulz: I got this path from a mac, where it contained a single app (Chrome stuff)... -> I do hope that it is the right folder for mac...

pypandoc/pandoc_download.py
```diff
@@ -0,0 +1,162 @@
...
+# and adding the URL here
+PANDOC_URLS = {
+    "win32": "https://github.com/jgm/pandoc/releases/download/1.16.0.2/pandoc-1.16.0.2-windows.msi",
+    "linux": "https://github.com/jgm/pandoc/releases/download/1.16.0.2/pandoc-1.16.0.2-1-amd64.deb",
+    "darwin": "https://github.com/jgm/pandoc/releases/download/1.16.0.2/pandoc-1.16.0.2-osx.pkg"
+}
+
+DEFAULT_TARGET_FOLDER = {
+    "win32": "~\\AppData\\Local\\Pandoc",
+    "linux": "~/bin",
```
JanSchulz: According to my [completely scientific twitter question](https://twitter.com/janschulz/status/698891984879144960) on `bin` vs `.bin`, this is 3x (3 votes vs 1 vote, highly significant...) more likely :-)

pypandoc/pandoc_download.py
```diff
@@ -0,0 +1,162 @@
...
+# Adding a new platform means implementing unpacking in "DownloadPandocCommand"
+# and adding the URL here
+PANDOC_URLS = {
+    "win32": "https://github.com/jgm/pandoc/releases/download/1.16.0.2/pandoc-1.16.0.2-windows.msi",
+    "linux": "https://github.com/jgm/pandoc/releases/download/1.16.0.2/pandoc-1.16.0.2-1-amd64.deb",
+    "darwin": "https://github.com/jgm/pandoc/releases/download/1.16.0.2/pandoc-1.16.0.2-osx.pkg"
+}
+
+DEFAULT_TARGET_FOLDER = {
+    "win32": "~\\AppData\\Local\\Pandoc",
```
JanSchulz: This is the default install location the installer chooses when you just click "next"...

bebraw:
> IMO there is only python setup.py bdist_wheel. What's the error when you run that? (I assume that you have wheel installed).

```bash
$ python setup.py bdist_wheel
forcing platform specific wheel name...
usage: setup.py [global_opts] cmd1 [cmd1_opts] [cmd2 [cmd2_opts] ...]
   or: setup.py --help [cmd1 cmd2 ...]
   or: setup.py --help-commands
   or: setup.py cmd --help

error: invalid command 'bdist_wheel'
```

I might be missing something obvious. I am a little out of touch with Python these days.

JanSchulz:
I think you are missing the wheel: `pip install wheel`.

[Sorry for the wrong `dist_wheel`, just now found it and fixed that above...]

bebraw:
Yeah, I made sure `wheel` is installed. I can see that `import wheel` part works. It's still giving `error: invalid command 'bdist_wheel'`, though. It feels like I'm missing something simple here. Just can't see what.

JanSchulz:
Do you use the same python for both the `python setup.py bdist_wheel` and `python -c "import wheel"`? setup.py tries to import wheel, but does not report the error: https://github.com/bebraw/pypandoc/pull/92/files#diff-2eeaed663bd0d25b7e608891384b7298R71
Could you change `pass` to `print("Not found")`?

bebraw:
> Do you use the same python for both the python setup.py bdist_wheel and python -c "import wheel"? 

This should match to system level Python (2.7). Latter doesn't error. And I'm sure former finds `wheel` as well.

Doing

```
        try:
            import wheel
            from distutils.version import StrictVersion
            if not StrictVersion(wheel.__version__) >= StrictVersion("0.27"):
                msg = "Including pandoc in wheel needs wheel >=0.27 but found %s.\nPlease update wheel!"
                raise RuntimeError(msg % wheel.__version__)
            print 'found'
        except ImportError:
            # No wheel installed, the user will get an error at a different place...
            print 'not found'
```

yields `found` at console.

JanSchulz:
Ok, I'm out of options. does `python setup.py download_pandoc install` work?

bebraw:
> Ok, I'm out of options. does python setup.py download_pandoc install work?

If I sudo it, yeah. It seems to write on system level so that seems to be required.

Maybe it would make sense to try to run the command through virtualenv or something?

bebraw:
@JanSchulz Are you sure it's not missing some code? I can see a line like

```python
cmd_classes = {'download_pandoc': DownloadPandocCommand}
```

As far as I can see, `setup` matches the command based on that. There's no similar entry for `bdist_wheel`, though.

JanSchulz:
bdist_wheel comes with the wheel package, so if that is installed, bdist_wheel should work. Up to now it was only included here as it overwrote the original command to implement the hack... As that isn't nessesary anymore, it's no in the cmd_classes anymore.

JanSchulz:
re virtualenv: virtualenv (and conda) usually just activate an environment by setting the python binary directory at the first place in PATH, so typing python will get you the python from the environment, not the system one. I've no clue about virtualenv but if you have a conda installation, you can create a env via `conda create -n <name> python=3.5 ...`

JanSchulz:
Anyway: if it was installed, `python -c "import pypandoc; pypandoc.get_pandoc_path()"` should now point to the pandoc in the python install directory (if you didn't have any newer pandoc in path)

JanSchulz:
And `python -c "import pypandoc; pypandoc.download_pandoc()"` should download pandoc to `~/bin` (linux) or `~/Applications/pandoc` (Mac). and these should be used afterwards (at least if you remove any other pandoc, e.g. the one in the python install directory under `Lib/site-packages/{,pypandoc*.egg/}pypandoc/files/`

bebraw:
Ok, I'll give it another go in a day or two. :+1:

bebraw:
> Does wheel building work? python setup.py download_pandoc bdist_wheel

I installed Python through brew and reinstalled wheel as well. Everything works now:

```bash
$ python setup.py bdist_wheel
found
forcing platform specific wheel name...
running bdist_wheel
running build
running build_py
creating build
creating build/lib
creating build/lib/pypandoc
```

> Does pypandoc.download_pandoc() work?

This works as well.

> Does it finds the installed pandoc?

How can I verify this?

Overall it's looking good. Go ahead with a release if you want. :+1:

JanSchulz:
> > Does it finds the installed pandoc?
>How can I verify this?

`python -c "import pypandoc; pypandoc.get_pandoc_path()"`

If you have a pandoc installed in path, you might need to remove that first, either by removing the element form PATH or by removing the binary... Or downgrading to a version which is older than the one which is downloaded (1.16.0.2) :-)

bebraw:
I set up a little script like this:

```python
import pypandoc

print pypandoc.get_pandoc_path()
```

Executing it yields:

```bash
$ python down.py
pandoc
```

So looks like it's pointing to the system pandoc in my case. Is that fine?

JanSchulz:
> So looks like it's pointing to the system pandoc in my case. Is that fine?

That's fine, but doesn't test whether the installer also could pick up the newly installed one. I think I will need to add a test to travis for this... Lets see...

JanSchulz:
Damn:

```
ERROR: test_conversion_with_citeproc_filter (__main__.TestPypandoc)
  File "tests.py", line 161, in test_conversion_with_citeproc_filter
    outputfile=None, filters=filters)
  File "/home/travis/build/bebraw/pypandoc/pypandoc/__init__.py", line 53, in convert
  File "/home/travis/build/bebraw/pypandoc/pypandoc/__init__.py", line 94, in _convert
  File "/home/travis/build/bebraw/pypandoc/pypandoc/__init__.py", line 178, in _process_file
    'Pandoc died with exitcode "%s" during conversion: %s' % (p.returncode, stderr)
RuntimeError: Pandoc died with exitcode "85" during conversion: pandoc: Filter pandoc-citeproc not found
```

It seems pandoc does not search for pandoc-citeproc in the directory where pandoc itself is and therefore can't find it when we install it into the normal location :-( Seems that we have to manipulate the PATH to include that dir... Tomorrow...

JanSchulz:
Finally green again...

@bebraw could you give this a test on macosx (AFAIK you are running OSX?) and if this works please merge...

bebraw:
@JanSchulz Sure. I'll try to get that done today. Any specific commands to run? I'll report back based on that. :+1:

JanSchulz:
If you really want to do something, remove pandoc from your path and then run `python setup.py donwload_pandoc; python tests.py`...

bebraw:
I removed pandoc from my system altogether. Here's output based on your commands:

```bash
pypandoc * [overhaul] $ python ./setup.py download_pandoc
Maybe try:

    brew install pandoc
See http://johnmacfarlane.net/pandoc/installing.html
for installation options
---------------------------------------------------------------



!!! pandoc not found, long_description is bad, don't upload this to PyPI !!!


running download_pandoc
* Using already downloaded file pandoc-1.16.0.2-osx.pkg
* Unpacking pandoc-1.16.0.2-osx.pkg to tempfolder...
x .
x ./usr
x ./usr/local
x ./usr/local/bin
x ./usr/local/bin/pandoc
x ./usr/local/bin/pandoc-citeproc
x ./usr/local/share
x ./usr/local/share/man
x ./usr/local/share/man/man1
x ./usr/local/share/man/man1/pandoc-citeproc.1
x ./usr/local/share/man/man1/pandoc.1
* Copying pandoc to .../pypandoc/files ...
* Copying pandoc-citeproc to .../pypandoc/files ...
* Done.
```

```bash
pypandoc * [overhaul] $ python -c "import pypandoc; print(pypandoc.get_pandoc_version())";
Maybe try:

    brew install pandoc
See http://johnmacfarlane.net/pandoc/installing.html
for installation options
---------------------------------------------------------------

  File "<string>", line 1, in <module>
  File "pypandoc/__init__.py", line 254, in get_pandoc_version
    _ensure_pandoc_path()
  File "pypandoc/__init__.py", line 360, in _ensure_pandoc_path
    raise OSError("No pandoc was found: either install pandoc and add it\n"
OSError: No pandoc was found: either install pandoc and add it
to your PATH or or call pypandoc.download_pandoc(...) or
install pypandoc wheels with included pandoc.
```

Any idea why the latter one is failing? I might be doing something stupid here.

JanSchulz:
@bebraw No real clue: can you uncomment the debug output in `pypandoc/__init__.py`  `_ensure_pandoc_path` (there are a few prints...)

JanSchulz:
But something is fishy here:

```
* Copying pandoc to .../pypandoc/files ...
* Copying pandoc-citeproc to .../pypandoc/files ...
```

This should also output that it made both things executable. -> https://github.com/bebraw/pypandoc/pull/92/files#diff-5d5106ab2bc99eacc6d82b01f674f9a3R93

Did you update the git content?

bebraw:
@JanSchulz Yeah. I set your repo as a remote and fetched the branch. All ok now.

JanSchulz:
But not on travis:

```
Trying: /home/travis/build/bebraw/pypandoc/pypandoc/files/pandoc
pandoc: HOME: getAppUserDataDirectory: does not exist (no environment variable)
Couldn't call pandoc to get version information. Output from pandoc:
('', None)
```

WTF?

JanSchulz:
So, the travis problem was that tox removes HOME and pandoc does not like that. In this case it resulted in different behaviour when running the same command in tox and outside... I now add HOME to the vars which ox should pass on...

I also added a print to stderr when we find an existing pandoc but the pandoc is not used... Lets see if this gets us any complains...

JanSchulz:
IMO finally ready to merge... @bebraw please do if you find everything to your satisfaction...

I would like to do a release based on this. The conda-forge repository now also has a pandoc package, so that means I can point people there and everything will work out of the box :-)

bebraw:
@JanSchulz Feel free to go ahead with the release. Thanks for the effort!

Issue #94, opened 2016-05-07
Incorrect path in appened to pandoc search_paths


pafcu:
pypandoc attempts to search for /usr/bin/pandoc, but accidentally just adds /usr/bin to the search_paths in the _ensure_pandoc_path() function. The problematic line is https://github.com/bebraw/pypandoc/blob/master/pypandoc/__init__.py#L310.

This is only a big problem if /usr/bin is not in your PATH, but it does result in some diagnostic error messages being printed (since subprocess.Popen attempts to execute a directory).

The solution is to further append the binary name ('pandoc') to the path.

janschulz:
Thanks, fix is incoming...

PR #95, opened 2016-05-07
Bugfix for search path and update to newer pandoc version

janschulz:
The search path in one case was missing the binary name in the end (`pandoc`).

Fixes: https://github.com/bebraw/pypandoc/issues/94


janschulz:
travis passes: merging

Issue #86, opened 2015-12-13
pypandoc.convert() doesn't handle errors properly


marbu:
Function `pypandoc.convert()` doesn't handle errors properly.

When I call `pypandoc.convert(filepath, 'rst')` with filepath which doesn't
exists, I got puzzling exception `RuntimeError: Missing format!` insdead of
statement of actual root cause of the error.

Similar issue happen when an empty string or int is passed as a filepath.

## Reproducer

Here is simple reproducer:

```
#!/usr/bin/env python
# -*- coding: utf8 -*-


import traceback
import sys

import pypandoc


files = [
    '/home/martin/projects/pyp2rpm/README.md',
    '/home/martin/projects/pypandoc/README.md',
    '/home/martin/tmp/doesnotexists.md',
    '',
    42,
    None,
    ]

for filepath in files:
    print("converting '{0}' (type: {1})".format(filepath, type(filepath)))
    try:
        pypandoc.convert(filepath, 'rst')
    except Exception:
        traceback.print_exc(file=sys.stdout)
    print()
```

When I run this script on my machine where 1st two files exists, I got
the following output:

```
$ ./reproducer.py
converting '/home/martin/projects/pyp2rpm/README.md' (type: <class 'str'>)

converting '/home/martin/projects/pypandoc/README.md' (type: <class 'str'>)

converting '/home/martin/tmp/doesnotexists.md' (type: <class 'str'>)
Traceback (most recent call last):
  File "./reproducer.py", line 23, in <module>
    pypandoc.convert(filepath, 'rst')
  File "/home/martin/projects/pypandoc/.env/lib/python3.4/site-packages/pypandoc-1.1.2-py3.4.egg/pypandoc/__init__.py", line 50, in convert
    outputfile=outputfile, filters=filters)
  File "/home/martin/projects/pypandoc/.env/lib/python3.4/site-packages/pypandoc-1.1.2-py3.4.egg/pypandoc/__init__.py", line 68, in _convert
    raise RuntimeError('Missing format!')
RuntimeError: Missing format!

converting '' (type: <class 'str'>)
Traceback (most recent call last):
  File "./reproducer.py", line 23, in <module>
    pypandoc.convert(filepath, 'rst')
  File "/home/martin/projects/pypandoc/.env/lib/python3.4/site-packages/pypandoc-1.1.2-py3.4.egg/pypandoc/__init__.py", line 50, in convert
    outputfile=outputfile, filters=filters)
[...]

PR #96, opened 2016-05-07
New main api: convert_file and convert_text

janschulz:
This introduces `convert_text()` and `convert_file()` as a successor to `convert()`. It also adds a few tweaks to the `convert()` path, so that it will give better error messages (more or less for https://github.com/bebraw/pypandoc/issues/86), but the real fix is to convert `convert()` calls into the new  `conver_*()` ones...

Also cleans up the old separation of the functions which seems to be only used by the tests, reactors tests and enables more tests on windows, adds tests for the new ways to use the library and updates the readme (including a new source where you can get the conda packages)

This closes: https://github.com/bebraw/pypandoc/issues/86

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,43 +9,64 @@ document converter.
 
 ## Installation
 
-pypandoc uses pandoc, so it needs an available installation of pandoc. For some common cases
-(wheels, conda packages), pypandoc already includes pandoc (and pandoc_citeproc) in it's
+pypandoc uses `pandoc`, so it needs an available installation of `pandoc`. For some common cases
+(wheels, conda packages), pypandoc already includes `pandoc` (and `pandoc_citeproc`) in it's
 prebuilt package.
 
-If pandoc is already installed (`pandoc` is in the PATH), pypandoc uses the version with the
+If `pandoc` is already installed (`pandoc` is in the PATH), `pypandoc` uses the version with the
 higher version number and if both are the same, the already installed version. See [Specifying the location of pandoc binaries](#specifying_binaries) for more.
 
-To use pandoc filters, you must have the relevant filter installed on your machine.
+To use `pandoc` filters, you must have the relevant filter installed on your machine.
 
 ### Installing via pip
 
-Install via `pip install pypandoc`
+Install via `pip install pypandoc`.
 
 Prebuilt [wheels for Windows and Mac OS X](https://pypi.python.org/pypi/pypandoc/) include
 pandoc. If there is no prebuilt binary available, you have to
-[install pandoc yourself](#installing-pandoc).
+[install `pandoc` yourself](#installing-pandoc).
 
 If you use Linux and have [your own wheelhouse](http://wheel.readthedocs.org/en/latest/#usage),
-you can build a wheel which includes pandoc with
+you can build a wheel which include `pandoc` with
 `python setup.py download_pandoc; python setup.py bdist_wheel`. Be aware that this works only
 on 64bit intel systems, as we only download it from the
 [official source](https://github.com/jgm/pandoc/releases).
 
 ### Installing via conda
 
-Install via `conda install -c https://conda.anaconda.org/janschulz pypandoc`.
+`pypandoc` is included in [conda-forge](https://conda-forge.github.io/). The conda packages will
+also install the `pandoc` package, so `pandoc` is available in the installation.
+
+Install via `conda install -c conda-forge pypandoc`.
 
 You can also add the channel to your conda config via
-`conda config --add channels https://conda.anaconda.org/janschulz`. This makes it possible to
+`conda config --add channels conda-forge`. This makes it possible to
 use `conda install pypandoc` directly and also lets you update via `conda update pypandoc`.
-
-Conda packages include pandoc and are available for py2.7, py3.4 and py3.5,
-for Windows (32bit and 64bit), Mac OS X (64bit) and Linux (64bit).
 
 ### Installing pandoc
 
-pandoc is available for many different platforms:
+If you don't get `pandoc` installed via a prebuild wheel which includes `pandoc` or via the
+conda package dependencies, you need to install `pandoc` by yourself.
+
+#### Installing pandoc via pypandoc
+
+Installing via pypandoc is possible on Windows, Mac OS X or Linux (Intel-based):
+
+```python
+# expects a installed pypandoc: pip install pypandoc
+from pypandoc.pandoc_download import download_pandoc
+# see the documentation how to customize the installation path
+# but be aware that you then need to include it in the PATH
+download_pandoc()
+```
+
+The default install location is included in the search path for `pandoc`, so you
+don't need to add it to `PATH`.
+
+#### Installing pandoc manually
+
+Installing manually via the system mechanismen is also possible. Such installation mechanismen
+make `pandoc` available on many more platforms:
 
 - Ubuntu/Debian: `sudo apt-get install pandoc`
 - Fedora/Red Hat: `sudo yum install pandoc`
@@ -56,6 +77,10 @@ pandoc is available for many different platforms:
   [here](http://johnmacfarlane.net/pandoc/installing.html)
 - [FreeBSD port](http://www.freshports.org/textproc/pandoc/)
   - Or see http://johnmacfarlane.net/pandoc/installing.html
+
+Be aware that not all install mechanismen put `pandoc` in `PATH`, so you either
+have to change `PATH` yourself or set the full path to `pandoc` in
+`PYPANDOC_PANDOC`. See the next section for more information.
 
 ### <a name="specifying_binaries"></a>Specifying the location of pandoc binaries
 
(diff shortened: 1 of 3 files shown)


janschulz:
@bebraw Could you have a look over this, I don't want to merge this without a nice review, this is a bit bigger... :-/

bebraw:
Sure, I'll have a look. :+1:

README.md
```diff
@@ -36,18 +36,31 @@ on 64bit intel systems, as we only download it from the
...
-Conda packages include pandoc and are available for py2.7, py3.4 and py3.5,
-for Windows (32bit and 64bit), Mac OS X (64bit) and Linux (64bit).
-
 ### Installing pandoc
 
-pandoc is available for many different platforms:
+If you don't get pandoc installed via the wheels which include pandoc or via the
+conda package dependencies, you need to install pandoc by yourself.
+
+Installing it on Windows, Mac OS X or Linux (Intel-based) is possible via:
```
bebraw: The content splits here in a weird way. Perhaps change the order or wording.

README.md
```diff
@@ -59,42 +72,44 @@ pandoc is available for many different platforms:
 - [FreeBSD port](http://www.freshports.org/textproc/pandoc/)
   - Or see http://johnmacfarlane.net/pandoc/installing.html
 
+If pandoc is in `PATH` or the default install location chosen by `download_pandoc()` it should
```
bebraw: Probably *in the `PATH`* though without `the` might work (you would have to look that up from somewhere).

README.md
```diff
@@ -59,42 +72,44 @@ pandoc is available for many different platforms:
...
+be picked up by `pypandoc`.
+
 ## Usage
 
-The basic invocation looks like this: `pypandoc.convert('input', 'output format')`. `pypandoc`
-tries to infer the type of the input automatically. If it's a file, it will load it. In case you
-pass a string, you can define the `format` using the parameter. The example below should clarify
-the usage:
+There are two basic ways to use pypandoc: with input files or with input strings. Using files
+will infer the input format from the filename.
```
bebraw: It is possible to force input format still, though? If so, maybe that's good to mention. It's difficult to rely on extensions (might be wrong for some reason). This would give a consistent API so you could pass an optional `format='md'` kind of parameter.

README.md
```diff
@@ -59,42 +72,44 @@ pandoc is available for many different platforms:
...
 It's also possible to directly let pandoc write the output to a file. This is the only way to
-convert to some output formats (e.g. odt, docx, epub, epub3, pdf). In that case `convert()` will
+convert to some output formats (e.g. odt, docx, epub, epub3, pdf). In that case `convert_*()` will
 return an empty string.
 
 ```python
 import pypandoc
 
-output = pypandoc.convert('somefile.md', 'docx', outputfile="somefile.docx")
+output = pypandoc.convert_file('somefile.md', 'docx', outputfile="somefile.docx")
```
bebraw: I wonder if we should just spell out `to` format so you would get something like this:

```python
output = pypandoc.convert_file('somefile.md', to='rst') # accepts optional from
output = pypandoc.convert_text('#some title', from='rst', to='md')
```

So instead of having `format`, we would have named arguments `to` and `from`. That reads well to me (no need to guess or look up the API).

README.md
```diff
@@ -112,53 +127,50 @@ pypandoc now supports easy addition of
...
+                         extra_args=pdoc_args,
+                         filters=filters)
 ```
-Please pass any filters in as a list and not a string.
+Please pass any filters in as a list and not as a string.
 
 Please refer to `pandoc -h` and the
 [official documentation](http://johnmacfarlane.net/pandoc/README.html) for further details.
 
+Note: the old way of using `convert(input, output)` is deprecated as in some cases it wasn't
```
bebraw: You could prefix this with `>`. Better formatting. One option would be to push this note to the change log though it can live here for a while.

README.md
```diff
@@ -112,53 +127,50 @@ pypandoc now supports easy addition of
...
 
 ```python
-output = pypandoc.convert('demo.md', 'pdf', outputfile='demo.pdf',
+output = pypandoc.convert_file('demo.md', 'pdf', outputfile='demo.pdf',
   extra_args=['-V', 'geometry:margin=1.5cm'])
 ```
 
-Note that it's important to separate `-V` and its argument within a list like that or else it won't work. This gotcha has to do with the way `subprocess.Popen` works.
+Note that it's important to separate `-V` and its argument within a list like that or else
+it won't work. This gotcha has to do with the way `subprocess.Popen` works.
```
bebraw: It would be good to link to `subprocess.Popen` here so the user can understand it better.

README.md
```diff
@@ -112,53 +127,50 @@ pypandoc now supports easy addition of
...
 
 ## Getting Pandoc Version
 
-As it can be useful sometimes to check what Pandoc version is available at your system, `pypandoc` provides an utility for this. Example:
+As it can be useful sometimes to check what Pandoc version is available at your system or which
+particular pandoc binary is used by `pypandoc`. For that, `pypandoc` provides teh following
```
bebraw: We could agree on either "pandoc" or "`pandoc`" and use that consistently. No idea which one is preferable, though. I might go with the latter myself.

pypandoc/__init__.py
```diff
@@ -14,8 +14,9 @@
 __author__ = u'Juho Vepsäläinen'
```
bebraw: If it's possible to set multiple authors, please do so. 👍

bebraw:
I did first pass on the PR. I don't have much to say on the code (too out of touch with Python) but having that test suite around helps for sure. Release-wise it would be a good idea to run 2.0 betas before pushing out the final so people hopefully have time to test the changes and provide feedback.

Overall this is a good step to make. 👍

README.md
janschulz: I think "in the path" refers to a physical path, but "in PATH" to "in the variable PATH". That's at least what google tells whe when I search for these two phrases..

README.md
janschulz: done...

README.md
janschulz: added another example to clarify.

README.md
janschulz: You mean changing the new API to `convert_xxx(input, to, from=None,...)? Unfortunately, you can't have "named arguments only" until py3, so a more natural `convert_xxx(input, from=None, to= None)` is IMO out as you would always need to specify the second parameter and can't write `convert_xxx("input.md", 'rst')` :-(

I will add a `from` argument and soft deprecate `format`.

README.md
bebraw: Ok, that's cool then.

README.md
bebraw: Do what feels the right solution to you. I've started getting used to ES6 (JavaScript) so I forgot about that limitation. 👍

janschulz: ok, from doesn't work, as `from` is a python name (`from x import y`). So format will stay...

bebraw: Interesting! Well, that's good to know. 👍

README.md
janschulz: done

README.md
janschulz: done: `pandoc` it is

pypandoc/__init__.py
janschulz: Not sure, I think it would be easier to change that to `pypandoc authors`.

Will do that in additional PR ...

janschulz:
Updated all README comments and also deprecated `convert()`.

pypandoc/__init__.py
bebraw: Yeah. Additional PR is just fine.

janschulz:
Not sure how to do betas: do you have any recommendations?

bebraw:
It's hard to find anything definitive about that. I can see versions like [Doorstop 0.7-beta.1](https://pypi.python.org/pypi/Doorstop/0.7-beta.1) at PyPI. Maybe the same could be done here. You would end up with `2.0-beta.1`. Ping our key users (active guys here) when publishing.

janschulz:
Ok, at least the technical side seems to be only releasing a `x.x.xb1` version: http://python-packaging-user-guide.readthedocs.io/en/latest/distributing/#pre-release-versioning

Do you have any mailinglists where you advertised pypandoc releases?

bebraw:
> Do you have any mailinglists where you advertised pypandoc releases?

Nope. I can do a post through my blog though so it will get to the Python blog aggregate (Planet Python). Just let me know when. We can also CC certain people here so they get the message.

bebraw:
@janschulz Do you want to proceed with this? 👍

janschulz:
Sorry, I more or less forgot about this PR... I tried to setup a new way to build the packages but up to now haven't done that :-(

Merging this now and then see what's still needed...

bebraw:
Cool. Let me know if you need anything. There's not much I can do codewise, but rest is possible. I can get the release on Planet Python etc.

janschulz:
I merged this without the fix for https://github.com/bebraw/pypandoc/issues/100 because the current fix was incomplete... Will post a new PR with a complete fix...

pypandoc/__init__.py
```diff
@@ -48,15 +50,135 @@ def convert(source, to, format=None, extra_args=(), encoding='utf-8',
...
+    warnings.warn(msg, DeprecationWarning, stacklevel=2)
+
+    path = _identify_path(source)
+    if path:
+        format = _identify_format_from_path(source, format)
+        input_type = 'path'
+    else:
+        source = _as_unicode(source, encoding)
+        input_type = 'string'
+        if not format:
```
manjitkumar: We could make `format` a required param by removing `format=none`.

bebraw: Can you open a separate issue for that? Thanks.

PR #99, opened 2016-05-15
Added section on specifying the location of pandoc binaries.

gwgundersen:


bebraw:
Looks good to me. 👍 

@janschulz Feel free to merge once you feel like it. 😄

janschulz:
@gwgundersen (Sorry for the late reply: Germany had some holidays...) There is already some sentence regarding the `PYPANDOC_PANDOC` right in the installation section: can you include this information in your new section and replace it in the installation section with a small sentence which includes a link to that section? That way it is all in one place :-)

gwgundersen:
@janschulz, good catch. I've consolidated the sentences into a single section. Please let me know if those sentences don't make sense together.

janschulz:
Thanks for pinging me, this looks great, thanks you!

PR #101, opened 2016-06-15
Fix for pdf conversion

### LINKED ISSUE #100: Pandoc died with exitcode "9" during conversion: b'pandoc: Unknown writer: pdf
To be honest I have no idea where exactly is the problem here.  I have following piece of code:

```python
import pypandoc

def main():
    parser = argparse.ArgumentParser()
    parser.add_argument("--fmt")
    parser.add_argument("--fail", dest="fail", action="store_true")
    parser.set_defaults(fail=False)
    args = parser.parse_args()

    if args.fail:
        to = args.fmt 
    else:
        # Ridiculous but works  
        to = "pdf" if args.fmt == "pdf" else args.fmt 

    pypandoc.convert(
        source="# Foo", to=to,
        format="markdown",
        outputfile="foo.pdf"
    )

if __name__ == "__main__":
    main()
```

which works just fine when I run it with `python ex2.py --fmt pdf` and let it use completely meaningless `"pdf" if args.fmt == "pdf" else args.fmt` but fails otherwise with 

```
RuntimeError: Pandoc died with exitcode "9" during conversion: b'pandoc: Unknown writer: pdf
To create a pdf with pandoc, use the latex or beamer writer and specify
an output file with .pdf extension (pandoc -t latex -o filename.pdf).'
```

Environment:

- Debian GNU/Linux
- Pandoc Version: 1.16.0.2~dfsg-1
- pypandoc==1.1.3
- Python  3.5.1, Python 2.7.11

janschulz:
converting to PDF needs pandoc to be invoked as `--to=latex -o something.pdf`.
Unfortunately we didn't enforce that:

* The old tests for pdf relied on `"pdf" is "pdf"` tests which is not always
  true of one of the strings was not interned.
* using the `convert(.., to="pdf")` was actually doing a conversion to html

The new checks ensure that, if `to="pdf"`:
* an outfile of type pdf is also given
* the actual invocation of pandoc passes `latex` as the to format

Fixes #100

### DIFF
--- a/pypandoc/__init__.py
+++ b/pypandoc/__init__.py
@@ -179,15 +179,20 @@ def _identify_input_type(source, format, encoding='utf-8'):
 
 
 def _validate_formats(format, to, outputfile):
-    formats = {
-        'dbk': 'docbook',
-        'md': 'markdown',
-        'rest': 'rst',
-        'tex': 'latex',
-    }
+    def normalize_format(fmt):
+        formats = {
+            'dbk': 'docbook',
+            'md': 'markdown',
+            'tex': 'latex',
+        }
+        fmt = formats.get(fmt, fmt)
+        # rst format can have extensions
+        if fmt[:4] == "rest":
+            fmt = "rst"+fmt[4:]
+        return fmt
 
-    format = formats.get(format, format)
-    to = formats.get(to, to)
+    format = normalize_format(format)
+    to = normalize_format(to)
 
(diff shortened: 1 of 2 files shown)


janschulz:
@bebraw Could you have a look and merge if satisfied?

bebraw:
I guess the next step might be to get a beta out. 👍

Issue #103, opened 2016-07-11
Pypandoc can not be used if $HOME is not set


tkob:
Pandoc and any functions of pypandoc fail is $HOME environment variable is not set:

```
>>> import os, pypandoc
>>> os.unsetenv('HOME')
>>> pypandoc.get_pandoc_version()
pandoc: HOME: getAppUserDataDirectory: does not exist (no environment variable)
Maybe try:

    sudo yum install pandoc
See http://johnmacfarlane.net/pandoc/installing.html
for installation options
---------------------------------------------------------------

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python2.7/site-packages/pypandoc/__init__.py", line 251, in get_pandoc_version
    _ensure_pandoc_path()
  File "/usr/lib/python2.7/site-packages/pypandoc/__init__.py", line 336, in _ensure_pandoc_path
    raise OSError("No pandoc was found: either install pandoc and add it\n"
OSError: No pandoc was found: either install pandoc and add it
to your PATH or install pypandoc wheels with included pandoc.
```

As the first line of the output, which comes from stderr, shows, `pandoc --version` tries to look up $HOME (to show default data directory) and fails because it is not set.

This is undesirable when pypandoc is used from a user with no $HOME, such as `apache` user.

A workaround is to use `os.putenv`, but the effect is somewhat non-local. Better solution would to utlilize `--data-dir` option of  `pandoc` command.

bebraw:
A `--data-dir` based PR would be welcome. Ideally the change should be backwards compatible.

janschulz:
Closed in #104

PR #104, opened 2016-07-11
Ensure $HOME is set before calling `pandoc --version`

tkob:
This PR fixes the problem described in #103 .

In #103, I suggested using `--data-dir` option, but it is not usable since pandoc ignores `--data-dir` option if `--version` is specified.

This PR takes another aproach; creates an environment which is local to `pandoc --version` and supply $HOME if it is not set.

### DIFF
--- a/pypandoc/__init__.py
+++ b/pypandoc/__init__.py
@@ -7,6 +7,7 @@ import textwrap
 import os
 import re
 import warnings
+import tempfile
 
 from .py3compat import string_types, cast_bytes, cast_unicode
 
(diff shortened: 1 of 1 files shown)


bebraw:
Looks good to me.

@janschulz Any thoughts? Want to merge and publish?

janschulz:
LGTM...

The problem I have right now is that I don't have my package builder running. I can cut a "normal" release without a problem, but not the wheels with embedded pandoc. 

So:
* normal source release is possible
* conda packages (via conda-forge) are possible
* wheels with pandoc included are not

bebraw:
Yeah, that's fine. I guess the embedded release can wait. Would it be possible to push that somewhere so it's easier to handle?

janschulz:
I had them build with the conda packages for my pwn channel, but doesn't work anymore :-( I want to have a similar setup: three CI services which build the wheels and upload them. Still waiting for some twine changes and then I have to sit down and write the CI scripts. But the latter seems to take a bit of time... A bit more probably, as debugging is hard with CI scripts :-/

janschulz:
Btw @tkob Does converting work? Or does that also need the same workaround?

tkob:
I believe `convert()` etc. now work since `pandoc` (i.e. no `--version`) does not fail when `HOME` is not set:
```
$ unset HOME
$ echo hello | pandoc
<p>hello</p>
$ pandoc --version
pandoc: HOME: getAppUserDataDirectory: does not exist (no environment variable)
```

janschulz:
@tkob Thanks!

Looks like this was missed upstream? Maybe file a bug there?

janschulz:
@bebraw I've cut a new release (1.2.0) and uploaded the sdist and updated the conda packages.

bebraw:
Nice. 👍

PR #105, opened 2016-07-13
Readme fix and version back to dev

janschulz:


bebraw:
Looks good to me.

janschulz:
Confirmed that the link works now and merged

Issue #112, opened 2016-10-27
Pandoc 1.18 compatibility


TomAugspurger:
Looks like the first change [here](http://pandoc.org/releases.html) broke some things for pypandoc.

Without looking at anything other than the error message, it looks like before you were parsing the output of `pandoc --version` to get the list of input and output formats. Now you should just be able to parse `pandoc --list-input-formats`, likewise for output.

bebraw:
Yeah, PR would be welcome. Ideally it should try to parse from `--list-input-formats` and if that fails, from `--version`.

janschulz:
Please see https://github.com/bebraw/pypandoc/pull/113 for a fix

PR #113, opened 2016-10-29
Fix for pandoc 1.18

janschulz:
pandoc 1.18 changed the way it lists input and output versions. We now try the
newer version first and if that fails the older version.

Tested on windows with 1.17.2 and 1.18

Closes: https://github.com/bebraw/pypandoc/issues/112


TomAugspurger:
Looks good, thanks!

Issue #125, opened 2016-12-29
pandoc_download should check architecture


zachriggle:
The code for `pandoc_download` does not check the current architecture.

This means that you can run `pandoc_download` on e.g. a Raspberry Pi, and it'll download the `amd64` distribution of Pandoc, and attempt to install it.

This ends up throwing a non-descriptive `RuntimeError` somewhere along the way.

Pandoc should check the current architecture in `pandoc_download`, and throw an exception (or similar) if a distribution is unavailable.

See Gallopsled/pwntools#820

ickc:
Some comments on this issue (not an expert here, so what I said might not make sense):

- the `pandoc_urls` format shouldn't need to be changed, because there's not other alternative `.deb` binaries provided so far, and it seems to me there's no plan to provide them in the future.

- to get the architecture, <http://stackoverflow.com/a/7171315> seems to help. But I don't have non-amd64 linux machine to test.

- It might be easier to catch the RuntimeError and emit error msg from there.

*If we want to improve the situation to provide support on platforms that pandoc don't provide a binary*:

- We can tell the system to install from existing package manager, e.g. apt-get on Raspberry Pi (which has a successful report in [pandoc-discuss/Install pandoc on Raspberry Pi ARM](https://groups.google.com/d/msg/pandoc-discuss/sGcUpLVmLnk/Qoh79_VhCQAJ)).

	- This seems manageable, but need to lookup different package manager from Linux variants. See [Pandoc - Installing pandoc](http://pandoc.org/installing.html#linux) for the supported Linux variants.

	- I see that pypandoc is available through conda too. Using system's package manager could be a problem in this case.

- If built from source (which I think is too complicated, but *might* fit conda better), references:

	- [Homebrew/homebrew-core/pandoc.rb](https://github.com/Homebrew/homebrew-core/blob/master/Formula/pandoc.rb) can be a good reference. (While homebrew is for macOS, there's a linux brew variants for linux. And the brew-formula could be learnt and converted into native Python instructions.)

	- (very old) For Raspberry Pi, see [Building pandoc on a Raspberry Pi 1 Model B · jgm/pandoc Wiki](https://github.com/jgm/pandoc/wiki/Building-pandoc-on-a-Raspberry-Pi-1-Model-B).

ickc:
Additional comment: I can see there's a need to check a matching architecture — before download rather than getting `RuntimeError` when trying installing — to avoid unnecessary download.

[...]

PR #126, opened 2017-01-14
Allow arbitrary pandoc version, default to "latest"

ickc:
See https://github.com/jgm/pandoc/releases/tag/1.19.1

### DIFF
--- a/pypandoc/pandoc_download.py
+++ b/pypandoc/pandoc_download.py
@@ -18,7 +18,7 @@ except ImportError:
 # Adding a new platform means implementing unpacking in "DownloadPandocCommand"
 # and adding the URL here
 # When updating pandoc version, update the following 2 variables
-INCLUDED_PANDOC_VERSION = "1.18"
+INCLUDED_PANDOC_VERSION = "1.19.1"
 DEB_SUBFFIX = "-1"
 
 URL_BASE = "https://github.com/jgm/pandoc/releases/download/" + INCLUDED_PANDOC_VERSION + "/pandoc-" + INCLUDED_PANDOC_VERSION
--- a/pypandoc/pandoc_download.py
+++ b/pypandoc/pandoc_download.py
@@ -7,6 +7,7 @@ import tempfile
 import os.path
 import subprocess
 import platform
+import re
 
 try:
     from urllib.request import urlopen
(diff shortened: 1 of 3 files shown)


ickc:
A related issue is to always use the latest version of pandoc, rather than hard-coding it in the code. An example of this in bash is in [ickc/pantable/.travis.yml](https://github.com/ickc/pantable/blob/master/.travis.yml#L55-L62). Basically it analyzes the redirected https://github.com/jgm/pandoc/releases/latest and find the binaries by regex (and version too if needed). This way, the packaging versioning from deb can also be obtained (which might not be `-1` in some cases).

If you think "pointing to the latest version of pandoc" is suitable for the needs of pypandoc, I can later make a separate pull request on this. (e.g. pandoc 2.0 is coming soon. It seems the only breaking change is related to pandoc as a Haskell library, and won't affect the cli. But some might not want to "accidentally" upgraded to 2.0.)

janschulz:
LGTM, will merge if you don#t want to implement the next stuff:

A "latest" approach (as a default) would be appreciated a lot. 

I would also like to see multiple version, so we could implement travis tests with a minimum pandoc version: put a version string argument into the download method which would try to pull that version (would probably also need a 'deb_version' argument and maybe some dict which would return the deb version for a regular pandoc version, so that a user wouldn't need to supply one).

pypandoc/pandoc_download.py
```diff
@@ -17,14 +17,19 @@
 # Uses sys.platform keys, but removes the 2 from linux2
 # Adding a new platform means implementing unpacking in "DownloadPandocCommand"
 # and adding the URL here
+# When updating pandoc version, update the following 2 variables
+INCLUDED_PANDOC_VERSION = "1.19.1"
+DEB_SUBFFIX = "-1"
+
+URL_BASE = "https://github.com/jgm/pandoc/releases/download/" + \
+    INCLUDED_PANDOC_VERSION + "/pandoc-" + INCLUDED_PANDOC_VERSION
+
```
janschulz: +1

janschulz:
Another related issue is https://github.com/bebraw/pypandoc/issues/125 which would need a change from the keys for the download URLs from `"linux"` to `("linux", "<arch>")`.

bebraw:
LGTM.

ickc:
Done:

pandoc_download.py: add _get_pandoc_urls to parse from HTML so that we can

- specify pandoc version as “latest” and parse from HTML automatically
- and need not to specify the deb_suffix but parse from HTML
automatically

ickc:
I didn't bump the version and will let you decide if this one should be bumped or wait.

ickc:
Sorry, I see the appveyor build is failing. Will check.

ickc:
Now fixed. I don't know why I typed 'wind32'. May be I don't like Windows in my gut ;)

more example:

```python
>>> pandoc_urls, version = pypandoc.pandoc_download._get_pandoc_urls("1.15.1"); print(pandoc_urls)
{'darwin': 'https://github.com/jgm/pandoc/releases/download/1.15.1/pandoc-1.15.1-osx.pkg', 'linux': 'https://github.com/jgm/pandoc/releases/download/1.15.1/pandoc-1.15.1-1-amd64.deb'}
>>> pandoc_urls, version = pypandoc.pandoc_download._get_pandoc_urls("1.16"); print(pandoc_urls)
{'win32': 'https://github.com/jgm/pandoc/releases/download/1.16/pandoc-1.16-1-windows.msi', 'linux': 'https://github.com/jgm/pandoc/releases/download/1.16/pandoc-1.16-3-amd64.deb', 'darwin': 'https://github.com/jgm/pandoc/releases/download/1.16/pandoc-1.16-1-osx.pkg'}
```

In the 1st case, win32 binary is not provided in 1.15.1 (see <https://github.com/jgm/pandoc/releases/tag/1.15.1>), so it won't even show up in the dict. (which would trigger this line `
    assert unpack is not None, "Can't handle download, only Linux, Windows and OS X are supported."`). In the certain sense this error msg can be rewritten to be more accurate but not very specific: `Can't handle download. Your platform is not supported by pandoc version...`.

In the 2nd case, note the deb-suffix is `-3`. The function knows it because the info is looked up from <https://github.com/jgm/pandoc/releases/tag/1.16> directly.

ickc:
I did just that: the assertion error msg becomes something like `"Can't handle download, Windows is not supported by pandoc 1.5.1."`

ickc:
> Another related issue is #125 which would need a change from the keys for the download URLs from "linux" to ("linux", "<arch>").

I didn't deal with issue #125 in this pull request. I'm going to put some comments there.

ickc:
> I did just that: the assertion error msg becomes something like "Can't handle download, Windows is not supported by pandoc 1.5.1."

I revert the last commit because it seems I might misunderstood the error msg I edited. If the message needs improvement, let's do it in a separate pull request.

**So the only change in this commit is how to obtain the urls to the binary and a new default "latest" version which always pointed to the latest pandoc version.**

ickc:
Or should I open another pull request, since this one was for a much simpler change?

pypandoc/pandoc_download.py
```diff
@@ -7,31 +7,62 @@
...
     "linux": "~/bin",
     "darwin": "~/Applications/pandoc"
 }
 
 
+def _get_pandoc_urls(version="latest"):
+    """Get the urls of pandoc's binaries
+    Uses sys.platform keys, but removes the 2 from linux2
+    Adding a new platform means implementing unpacking in "DownloadPandocCommand"
+    and adding the URL here
```
janschulz: This looks outdated? "adding the URL here"?

ickc: There's rummors (https://github.com/jgm/pandoc/issues/3317#issuecomment-272963420) that pandoc will add a `.zip` variant for the Windows binaries.

Anyhow, if in the future there's new releases other than `.deb|msi|pkg` then the "url" should be added in the regex below. Do you want me to improves the comments there?

PR #130, opened 2017-02-14
Fixed a crash with custom writers

kristoff3r:
When using a custom writer inside directories containing
either a + or - in the name, the file extension isn't
recognized properly because _get_base_format strips it
off. Can be triggered in the tests with:

mkdir /tmp/foo-bar/
TMPDIR=/tmp/foo-bar ./tests.py

### DIFF
--- a/pypandoc/__init__.py
+++ b/pypandoc/__init__.py
@@ -224,14 +224,14 @@ def _validate_formats(format, to, outputfile):
 
     base_to_format = _get_base_format(to)
 
-    file_extension = os.path.splitext(base_to_format)[1]
+    file_extension = os.path.splitext(to)[1]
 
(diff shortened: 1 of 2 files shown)


janschulz:
Can you add a test in test.py for that, so we don't reintroduce this bug?

janschulz:
Otherwise LGTM

kristoff3r:
I have added a test (or rather, I have made sure the customwriter test always triggers it).

Also, I found and fixed two bugs in the tests: 
- shutil.rmtree doesn't delete files, only directories
- Cleanup wasn't being performed when exceptions were triggered in PDF tests

janschulz:
Thanks a lot for your contribution!

Issue #134, opened 2017-05-10
pandoc_download leaves installer at working directory


sameersingh7:
The code does not remove the file from working directory.

So if code is run by Travis CI, the installer may get committed as well in the repository.
It should make sure that file is removed from the working directory after the run.

sameersingh7:
I'd like to work on this.

janschulz:
Do you have an example where this is happening?

The current sdist on pypi is fine: https://pypi.python.org/pypi/pypandoc

sameersingh7:
"Now, pypandoc leaves the installer in the current working directory after download, which is fine locally, but creates a problem when run on remote systems like Travis. The installer could get committed accidently to the repository. To solve this, I had to take a look at source code for pypandoc and call an internal method, which pypandoc basically uses to set the name of the installer. I use that method to find out the name of the file and then delete it after installation is over."

http://blog.fossasia.org/generating-a-documentation-site-from-markup-documents-with-sphinx-and-pandoc/

janschulz:
#135 is merged, so closing... I need to do a release...

PR #135, opened 2017-05-10
pandoc_download.py edited so that installer gets removed after pandoc installation.

sameersingh7:
Travis CI would commit the installer in the repository if not done so. #134

### DIFF
--- a/pypandoc/pandoc_download.py
+++ b/pypandoc/pandoc_download.py
@@ -205,3 +205,4 @@ def download_pandoc(url=None, targetfolder=None, version="latest"):
     assert unpack is not None, "Can't handle download, only Linux, Windows and OS X are supported."
 
     unpack(filename, targetfolder)
+    os.remove(filename)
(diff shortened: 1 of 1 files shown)


pypandoc/pandoc_download.py
```diff
@@ -205,3 +205,4 @@ def download_pandoc(url=None, targetfolder=None, version="latest"):
     assert unpack is not None, "Can't handle download, only Linux, Windows and OS X are supported."
 
     unpack(filename, targetfolder)
+    os.remove(filename)
```
sameersingh7: I think that leaving a choice whether to keep the installer or not to the user seems a more elegant way to go. Third option mentioned by you seems a pretty good choice.

sameersingh7:
@janschulz I have edited the file. Please review it.

janschulz:
I resolved the commits and will merge if Travis/appveyor are green

PR #137, opened 2017-05-15
get_pandoc_path: add option to download pandoc if not found.

xoviat:


janschulz:
Whats the usecase here? That code is mostly used internally and so you would need to explicitly call that method. Isn't it then easier to simply check and download in you own code?

I wouldn't mind a new API method like `ensure_pandoc_installed()` which either finds a pandoc OR downloads it sucessfuly OR raises an exception.

xoviat:
The use case is that of `ensure_pandoc_installed`. Is adding a new method the preferred option for that use case?

PR #145, opened 2017-09-10
Include name of source file path in error.

lawrennd:
Using pypandoc I get very confused with the error message "source_file is not a valid path": I call pypandoc recursively for my documents, so you can imagine it's hard to imagine where the error occurs!

Perhaps there is a good reason for not including the value of source_file in the message, but in my own local version, I find it very useful, so just submitting a pull request to include to (hopefully) make the error message clearer.

### DIFF
--- a/pypandoc/__init__.py
+++ b/pypandoc/__init__.py
@@ -134,7 +134,7 @@ def convert_file(source_file, to, format=None, extra_args=(), encoding='utf-8',
             path.
     """
     if not _identify_path(source_file):
-        raise RuntimeError("source_file is not a valid path")
+        raise RuntimeError(source_file + " is not a valid path")
     format = _identify_format_from_path(source_file, format)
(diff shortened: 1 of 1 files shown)


janschulz:
I love this, but this needs to be fixed before merging:
```
ERROR: test_call_with_nonexisting_file (__main__.TestPypandoc)
  File "tests.py", line 371, in test_call_with_nonexisting_file
    self.assertRaises(RuntimeError, f, filepath)
  File "C:\Python27-x64\lib\unittest\case.py", line 473, in assertRaises
    callableObj(*args, **kwargs)
  File "tests.py", line 368, in f
    pypandoc.convert_file(filepath, 'rst')
  File "C:\projects\pypandoc\pypandoc\__init__.py", line 137, in convert_file
    raise RuntimeError(source_file + " is not a valid path")
TypeError: unsupported operand type(s) for +: 'int' and 'str'
```

Issue #149, opened 2018-02-15
tests fail with 


keszybz:
[Another Fedora maintainer here]

I'm trying to build pypandoc with pandoc-2.0.6-1.fc28.x86_64, and unfortunately there are many test failures:
```
+ /usr/bin/python2 tests.py
test_basic_conversion_from_file (__main__.TestPypandoc) ... FAIL
test_basic_conversion_from_file_url (__main__.TestPypandoc) ... ok
test_basic_conversion_from_file_with_format (__main__.TestPypandoc) ... FAIL
test_basic_conversion_from_http_url (__main__.TestPypandoc) ... ERROR
test_basic_conversion_from_string (__main__.TestPypandoc) ... FAIL
test_basic_conversion_to_file (__main__.TestPypandoc) ... FAIL
test_call_with_nonexisting_file (__main__.TestPypandoc) ... ok
test_conversion_error (__main__.TestPypandoc) ... ok
test_conversion_from_markdown_with_extensions (__main__.TestPypandoc) ... FAIL
test_conversion_from_non_plain_text_file (__main__.TestPypandoc) ... FAIL
test_conversion_with_citeproc_filter (__main__.TestPypandoc) ... ok
test_conversion_with_empty_filter (__main__.TestPypandoc) ... ok
test_conversion_with_markdown_extensions (__main__.TestPypandoc) ... ok
test_convert_text_with_existing_file (__main__.TestPypandoc) ... ok
test_convert_with_custom_writer (__main__.TestPypandoc) ... ERROR
test_converts_valid_format (__main__.TestPypandoc) ... ok
test_depreaction_warnings (__main__.TestPypandoc) ... ok
test_does_not_convert_from_invalid_format (__main__.TestPypandoc) ... ok
test_does_not_convert_to_invalid_format (__main__.TestPypandoc) ... ok
test_get_pandoc_formats (__main__.TestPypandoc) ... ok
test_get_pandoc_path (__main__.TestPypandoc) ... ok
test_get_pandoc_version (__main__.TestPypandoc) ... FAIL
test_pdf_conversion (__main__.TestPypandoc) ... ok
test_unicode_input (__main__.TestPypandoc) ... ok

======================================================================
ERROR: test_basic_conversion_from_http_url (__main__.TestPypandoc)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "tests.py", line 174, in test_basic_conversion_from_http_url
    received = pypandoc.convert(url, 'html')
[...]

PR #156, opened 2018-04-05
Fix tests with newer pandoc

jankatins:
Closes: https://github.com/bebraw/pypandoc/issues/149


jankatins:
I've opened https://github.com/jgm/pandoc/issues/4610 for the uppercase lua filename problem.

PR #160, opened 2018-06-03
Added quiet to _ensure_pandoc_path and extra paths for Windows installs of Pandoc

burgerbecky:
Added the flag "quiet" to _ensure_pandoc_path to shut down output spew. enhanced ensure_pandoc_installed() to allow it to pass all parameters to download_pandoc() in case they need to be overridden


burgerbecky:
The second checkin adds support to scan for copies of pandoc that were installed into the "Program Files" or "Program Files(x86)" directories as they would be located if pandoc was manually installed using the msi installer for "all users".

jankatins:
Just for my curiosity: in what usecase was the output to much?

burgerbecky:
I have build tools that use a "verbose" flag to have it run in a quiet mode. When I added pypandoc, I wanted to have that ability to allow it to run without any text output if everything is working fine. I'm using the verbose flag in my build tools (Look at pyburger and makeprojects on github)

PR #164, opened 2018-09-14
Add verify_formats option to reduce calls to pandoc.

JarrahG:
This commit adds the verify_formats boolean flag to all convert_*
functions. This allows the caller to directly call pandoc rather than
having pypandoc run three subprocess calls before calling pandoc to
convert the text. This flag defaults to True, and should only be set
False when confident of the formats *and* performance is an issue.

For example, this was originally created to speed up compilation of a
document which consisted of over 10,000 html files to be converted to
latex. Significant improvements were found here, though obviously this isn't a good default.


JarrahG:
The latest commit is just a rebase on top of the current master branch. 
Thanks for having a look at this.

NicklasTegner:
hi @JarrahG 

Can you pull the latest commits from master into your branch, so we can make sure your code still works?

PR #181, opened 2019-06-05
Skip code that isn't needed when the url is set

ecederstrand:
This allows calling downloading the deb package from an internal server, when the build server doesn't otherwise have Internet access.


ecederstrand:
The test failures don't really look like something to do with this patch? `File `xcolor.sty' not found.`

ecederstrand:
Ping?

ecederstrand:
Pinging @burgerbecky, @jankatins, or @charlesreid1 Is anyone actively supporting this package?

charlesreid1:
🤷‍♂

Issue #155, opened 2018-04-05
Support pandoc v2 and --lua-filter


mdourmousoglou:
Currently all filters passed to `convert_input` are explicitly assumed to be JSON filters ([line 280](https://github.com/bebraw/pypandoc/blob/6d5a0a667c93220e6837db71525e738eade525c5/pypandoc/__init__.py#L280)). Pandoc v2 accepts Lua filters with a new `--lua-filter` option but it won't accept a Lua filter passed just with `--filter` (or the other way around).

slackline:
Is there scope for getting `pypandoc` to play ball with [`pandocfilters`](https://pypi.org/project/pandocfilters/)?  I know its not a Lua filter but if you're working with `pypandoc` in Python anyway.

jankatins:
Feel free to write PR for this. As I don't use filters, I cannot really play with this :-/

PR #194, opened 2020-06-23
Add support for lua filters.

jwidner:
Interpret filters with extension .lua as lua filters.
Closes: #155 

Explanation:
Ideally, we would be able to create separate arguments for filters and lua filters.
However, the order that filters appear as arguments to Pandoc actually matters
because it affects the order that the filters are applied to the AST.
Therefore, if a file's name ends in .lua in the filters list, it makes sense to
interpret it as a lua filter.


NicklasTegner:
hi @jwidner 
Can you pull the recent commits from master into your branch, so we can make sure your code still works as expected?

domvwt:
It seems this PR was closed due to inactivity. Would it be possible to reopen and complete the implementation? I'll be happy to take this on if it's not already taken care of.

PR #195, opened 2020-06-23
Fix #189: bug with download_pandoc() affecting Windows.

### LINKED ISSUE #189: Error when using pypandoc.download_pandoc() on windows
Hey, 

Thanks for `pypandoc` I have been using it for quite some while now ;). However, there is a tiny problem when using `pypandoc.download_pandoc()` on windows. The installer will be downloaded but cannot be executed due to a wrong path.

Steps to reproduce: 

Create the following folder structure 

	.
	├── app.py
	├── download_folder
	└── target_folder

Run 

~~~
# using python 3.7.5
pipenv --three 

pipenv install pypandoc
pipenv run python -c "import pypandoc; print(pypandoc.__version__)"
# 1.5 
~~~

Run the following code snippet

~~~py 
from pathlib import Path 

import pypandoc 
print(pypandoc.__version__)

target_folder = Path('./target_folder').resolve()
download_folder = Path('./download_folder').resolve()
pypandoc.download(targetfolder=str(target_folder), download_folder=str(download_folder))
~~~

The installer will be downloaded to `download_folder`. However, the installer path passed to `subprocess.check_call` is wong. The path to the installer is generated via: 

~~~py
# pypandoc/download_pandoc.py line 195
filename = os.path.expanduser(download_folder) + '/' + filename
~~~

So the generated path  on windows will be: `C:\\Some\\Path\\to/my_installer.msi` about which windows complains. 

Thank you for looking into this :).

jwidner:
Make the code that finds the installer less platform dependent by using the os.path module.
Closes: #189 

I confirmed the bug, and then tested the fix on Windows, and it seems to work.

### DIFF
--- a/pypandoc/pandoc_download.py
+++ b/pypandoc/pandoc_download.py
@@ -192,7 +192,7 @@ def download_pandoc(url=None, targetfolder=None, version="latest", quiet=False, 
         if download_folder.endswith('/'):
             download_folder = download_folder[:-1]
 
-        filename = os.path.expanduser(download_folder) + '/' + filename
+        filename = os.path.join(os.path.expanduser(download_folder), filename)
 
     if os.path.isfile(filename):
         print("* Using already downloaded file %s" % (filename))


sbland:
Any chance this could have caused issue #200 ?

NicklasTegner:
hi @jwidner 
Can you pull the latest commits in from the master branch, and then check if:

1. The bug is still there, and
2. if your code fixes it?

Thanks :)

PR #204, opened 2020-11-22
Fix tests for new heading format in pandoc 2.11.2 output

sternenseemann:
pandoc 2.11.2 outputs markdown headings using hashes (#) instead of
underlining them with =, - and the like. From the changelog:

> Default to using ATX (##-style) headings for Markdown output
> (#6662, Aner Lucero). Previously we used Setext (underlined)
> headings by default for levels 1–2.

<https://pandoc.org/releases.html#pandoc-2.11.2-2020-11-19>

With this fix tests work with both version before and after the change. Tested with:

* pandoc 2.11.2
* pandoc 2.11.1.1

There is also an alternative solution for this: Use h3, h4, … for which the format doesn't change, I believe. If you prefer that solution, I'd be happy to implement it.

### DIFF
--- a/tests.py
+++ b/tests.py
@@ -10,6 +10,7 @@ import sys
 import tempfile
 import unittest
 import warnings
+import functools
 
 import pypandoc
 from pypandoc.py3compat import path2url, string_types, unicode_type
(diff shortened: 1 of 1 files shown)


NicklasTegner:
hi @sternenseemann 

Could you try pulling in the latest commits from master and then see if your code still works?

sternenseemann:
I think this change is obsolete now since 143b57f5af230b89aee41f4b4c16a74be81d01b5 and ee5530a0183f330ac6b02640482fa7fbd48c36c1 removed the usage of headings in the test, thus making this change non applicable.

NicklasTegner:
Ok. In that case, I'll close this pr.
Thanks for your work!

PR #217, opened 2021-06-02
Prevent new console  window being created on Windows

LukeBriggsDev:
When using deployment methods such as pyinstaller, pypandoc would
flash open a console window when it invoked pandoc.

Changing the creation flag on windows to prevent a new console window
fixes this


NicklasTegner:
hey @LukeBriggsDev 

Can you pull the recent commits into your branch, so we can see if your code still works (the test cases should all pass now)

LukeBriggsDev:
All recent commits pulled and tests pass

NicklasTegner:
Thanks!
Good work with the fix

PR #225, opened 2021-07-24
BUG: Need to add targetfolder to PATH when ensuring installed

scottclowe:
We need to add targetfolder to the PATH environment variable otherwise we will not be able to run pandoc as a subprocess when it is housed this arbitrary download location.

This PR addresses issue as discussed in https://github.com/NicklasTegner/pypandoc/issues/97#issuecomment-886092575.


NicklasTegner:
Good one!
Thanks for your contribution :)

PR #227, opened 2021-08-04
Also encode stderr, for better display of errors

### LINKED ISSUE #226: Make error message in RuntimeException readable
When pandoc has an error, the output from pandoc gets put into the Runtime-Exception as bytes (directly from stderr). 

For me this results in the whole error-message from pandoc being in one line (even though it should have linebreaks). Especially when using "--verbose" this gets very unreadable.

I've checked and by encoding stderr in the same way as stdout in https://github.com/NicklasTegner/pypandoc/blob/master/pypandoc/__init__.py line 322 (_convert_input, towards the end), this error is fixed.

EorlBruder:
Fixes #226


NicklasTegner:
Good one. THanks for the contribution :)

PR #250, opened 2022-01-17
Issue all messages using the logging library

### LINKED ISSUE #249: Warnings issued on stderr are not reported
First, thanks for this super-useful package.

I was having some trouble with missing images in a newly created `.docx` file, but pypandoc was not reporting anything, so I decided to run the same arguments through pandoc directly. Running directly I got useful `[WARNING]` output, but on `stderr` not `stdout` which is why pypandoc does not report it. See the example below:

```Python console
>>> import sys
>>> import tempfile
>>> import subprocess
>>> from pathlib import Path
>>> import pypandoc
>>> tmpdir = tempfile.TemporaryDirectory()
>>> report_dir = Path(tmpdir.name)
>>> p = report_dir / "input.md"
>>> _ = p.write_text('![Mock](missing.png)\n')
>>> stdout = pypandoc.convert_file(
...     f"{p}",
...     'docx',
...     outputfile=f"{report_dir / 'input.docx'}",
...     extra_args=[f'--resource-path={report_dir}'])
>>> print(stdout)
<BLANKLINE>

>>> args = ['pandoc',
...         '--from=markdown',
...         '--to=docx',
...         f'{p}',
...         f'--output={report_dir / "input.docx"}',
...         '--sandbox',
...         '--resource-path={report_dir}']
>>> creation_flag = 0x08000000 if sys.platform == "win32" else 0
>>> p = subprocess.Popen(args,
...                      stdout=subprocess.PIPE,
...                      stderr=subprocess.PIPE,
...                      creationflags=creation_flag)
>>> stdout, stderr = p.communicate()
>>> print(stderr.decode('utf-8')) # doctest: +NORMALIZE_WHITESPACE
[WARNING] Could not fetch resource missing.png: PandocResourceNotFound "missing.png"
<BLANKLINE>
```

Would it be agreeable to modify `_convert_input` to return the output on both `stdout` and `stderr` rather than just `stdout`? Happy to put together a PR, if that would help.

Cheers,

Mat

EDIT: Here is my environment, in case it's useful:

```
> conda info

     active environment : _snld3d
    active env location : C:\Anaconda3\envs\_snld3d
            shell level : 2
       user config file : C:\Users\Work\.condarc
 populated config files : C:\Users\Work\.condarc
                          C:\Anaconda3\envs\_snld3d\.condarc
          conda version : 4.11.0
    conda-build version : 3.21.6
         python version : 3.9.7.final.0
       virtual packages : __win=0=0
                          __archspec=1=x86_64
       base environment : C:\Anaconda3  (writable)
      conda av data dir : C:\Anaconda3\etc\conda
  conda av metadata url : None
           channel URLs : https://conda.anaconda.org/conda-forge/win-64
                          https://conda.anaconda.org/conda-forge/noarch
                          https://repo.anaconda.com/pkgs/main/win-64
                          https://repo.anaconda.com/pkgs/main/noarch
                          https://repo.anaconda.com/pkgs/r/win-64
                          https://repo.anaconda.com/pkgs/r/noarch
                          https://repo.anaconda.com/pkgs/msys2/win-64
                          https://repo.anaconda.com/pkgs/msys2/noarch
          package cache : C:\Anaconda3\pkgs
                          C:\Users\Work\.conda\pkgs
                          C:\Users\Work\AppData\Local\conda\conda\pkgs
       envs directories : C:\Anaconda3\envs
                          C:\Users\Work\.conda\envs
                          C:\Users\Work\AppData\Local\conda\conda\envs
               platform : win-64
             user-agent : conda/4.11.0 requests/2.26.0 CPython/3.9.7 Windows/10 Windows/10.0.19043
          administrator : True
             netrc file : None
           offline mode : False
```

```
> conda list
...
pandoc                    2.16.2               h8ffe710_0    conda-forge
...
pypandoc                  1.7.2              pyhd8ed1ab_0    conda-forge
...
```

H0R5E:
This commit changes the issuing of messages to use the Python logging library. Logging in the module now follows this process:

1. The module logger is initiated on import (without a handler)
2. The user then has the opportunity to attach a handler to the logger before the pypandoc functions are called.
3. When a function that issues messages is called, check if the logger has a handler and if not add a console handler, formatted in the same manner as pandoc's.
4. When log messages are detected on Pandoc's stderr stream, parse them to determine the level and then reissue the log messages at the appropriate level.

Checking and setting up the logging handler is done by the `_check_log_handler` function (in the `handler` submodule), while determining the level of the messages issued by Pandoc on stderr is done by the `_classify_pandoc_logging` function.

Note, all public "quiet" flags are now deprecated, as this can now be handled by configuring the logger.

Closes #249


NicklasTegner:
amazing work @H0R5E 

I was wondering, if it would be best behavior to initialize a logger instead of just using print. Or would that be overkill.
An alternative could be to use the already imported [warmings module](https://docs.python.org/3/library/warnings.html).

What do you think?

H0R5E:
Thanks @NicklasTegner! I've been thinking about this PR and your suggestions.

Firstly, it struck me that the `quiet` flag is wrong! You can pass `--quiet` to pandoc in the `extra_args` to achieve the same thing, so I think the `quiet` flag I added should be removed.

Also, I don't think using the warnings module is the right approach because pandoc will put log messages with both INFO and WARNING level onto `stderr`.

Probably using the Python logging module is the most robust way to handle the `stderr` stream, but setting it up is a slightly complex process. Basically, I was considering this approach:

1. Initiate the module logger on import (without a handler)
2. The user then has the opportunity to attach a handler to the logger before the pypandoc functions are called. The logger could be muted by adding a null handler to it, for instance.
3. When `_convert_input` is called check if the logger has a handler and if not add a console handler, formatted in the same manner as pandoc's.
4. When log messages are detected on `stderr`, parse them to determine if they are at INFO or WARNING level and then reissue the log messages at the appropriate level.

The questions I have with this approach are "should the other functions that print (namely `_ensure_pandoc_path`) also use the logger, in which case also have responsibility for initializing it (and sharing the format?)?" and "is it just confusing to have this logging system working alongside the pandoc one?". My attempt at answers to these questions are that `_ensure_pandoc_path` should just keep printing and we use logging just to handle the pandoc output, and that if we do this right, the normal user won't notice that there is logging going on, while the power user has the opportunity to configure it.

So, in conclusion, I think the `quiet` flag should be removed either way as it is redundant. I think using the `logging` module is a good idea (to allow customization), but there are many ways to set it up and I wanted to check with you whether my idea above is sound.

Cheers,

Mat

NicklasTegner:
@H0R5E 
In that case, yes the quiet flag should be removed.

For logging, I would say that your list is a good solution, I would however put all printing (including from _ensure_pandoc_path) into the logging as well, at the appropriate levels.

H0R5E:
OK @NicklasTegner, I'll give it a try.

H0R5E:
@NicklasTegner, I've made the changes. Everything that printed or wrote to `stdout` or `stderr` now adds messages through the `pypandoc` logger. I've also removed **ALL** the `quiet` flags, including the existing ones. These didn't seem to make any sense anymore. For instance, `download_pandoc` simply muted stdout when the `quiet` flag was set, but I don't want to manipulate the logging level, because that is something under the user's control. Note that because I have removed the `quiet` flags this is a breaking change.

NicklasTegner:
@H0R5E  can you explain what exactly is a "breaking change" in this?

sthagen:
Usually any change in the public command line interface (removing a flag that users may have used and now they have to adapt) is a breaking change. But then, people usually embrace change because they want the beneficial parts that should come along with it - maybe just consider when bumping the semantic version.

NicklasTegner:
mmhmm. What do you guys think.

Either just remove the flags outright in this pr, bumping the version to 1.8 when relizing.
Or issuing a warning message, stating that the quiet flag is Deprecated, and will be removed in 1.8, where we will then be able to include this pr in the next minor release, like 1.7.x?

H0R5E:
So, to be precise, the breaking change is an API change to the `ensure_pandoc_installed` function:

```
-def ensure_pandoc_installed(url=None, targetfolder=None, version="latest", quiet=False, delete_installer=False):
+def ensure_pandoc_installed(url=None, 
+                            targetfolder=None,
+                            version="latest",
+                            delete_installer=False):
```

And an API change to the `download_pandoc` function:

```
-def download_pandoc(url=None, targetfolder=None, version="latest", quiet=False, delete_installer=False, download_folder=None):
+def download_pandoc(url=None,
+                    targetfolder=None,
+                    version="latest",
+                    delete_installer=False,
+                    download_folder=None):
```

My understanding is that semantic versioning would require that this then be a **major** version change (i.e. to version 2). 

Here are the options I see for finishing this:

1. Merge the PR as it is and bump to version 2. A major version change is indicative to users of breaking API changes.
2. Replace the `quiet` flags, but just with deprecation warnings and no functionality. Release this change as a minor release (i.e. 1.8) and then the deprecation warnings would be removed on the next major release (i.e. version 2).
3. Replace the existing quiet flags and re-implement their functionality by adding `if not quiet:` guards in front of all the relevant logging messages.

Happy to finish this off whichever way you want.

H0R5E:
I think option 2 in my list is my preference, BTW.

NicklasTegner:
I would say go ahead with option 2. Adding a message like:

> deprecation Warming: The quiet flag in PyPandoc has been deprecated in favor of logging. See LINK for more information

Just essentially a message, stating that we have removed the functionality in favor of logging with the logging module, and where to find more information e.g the readme.

H0R5E:
@NicklasTegner, I think this is ready now.

NicklasTegner:
fantastic work. Good job :) @H0R5E

PR #252, opened 2022-03-04
Add `cworkdir` kwarg to convert functions

### LINKED ISSUE #251: Set cwd for pandoc process
I'm building an API based around pypandoc and would like to have pandoc executing with the working directory set to a one-time-use temp directory. Is there a currently supported way to do that? From what I can tell the answer is no, and if that's the case, would you be open to a PR adding a way to do so?

mcm:
Add a `cworkdir` kwarg to all of the convert functions, which is passed as the `cwd` kwarg in `subprocess.Popen` when calling pandoc.

Note that `None` is the default for `cwd` in `subprocess.Popen`, so the default behavior is unchanged.

Let me know if anything else is needed.

Closes #251

### DIFF
--- a/pypandoc/__init__.py
+++ b/pypandoc/__init__.py
@@ -26,7 +26,7 @@ logger = logging.getLogger(__name__)
 
 
 def convert(source, to, format=None, extra_args=(), encoding='utf-8',
-            outputfile=None, filters=None):
+            outputfile=None, filters=None, cworkdir=None):
     """Converts given `source` from `format` to `to` (deprecated).
 
     :param str source: Unicode string or bytes or a file path (see encoding)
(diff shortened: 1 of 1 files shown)


sthagen:
In the appveyor runs all but python 3.6 failed (may need some digging as for me the details &rightarrow; "logs" only shows that the test execution failed (I think the pip ERROR further up the log is a red herring ... but may be good to find the cause)

mcm:
Hmm yeah that's interesting. It's like it's failing to even execute the tests or something.

NicklasTegner:
I'm gonna be honest here and say I have no idea why it's failing.
I see some complains about the poetry.lock file not being up-to-date, and some resolving of dependencies for twine is yelding warnings.
Nothing that should make the tests fail.

In addition, I also just ran it on my local Windows 10 box, with Python 3.9.7 and it also seems to be working fine.

mcm:
I took a look and wasn't really sure either. It's not even showing that the tests were run. Not sure if there's a way to resubmit and see if it was just a fluke.

mcm:
Hmm, looks like this time 3.6 timed out, and then the rest still failed.

NicklasTegner:
@mcm  from what I've been able to gather, it seems like a permission error in the CI environments.
Would you mind trying to change the code to something like the following
* Remove the cwd argument in the subprocess calls

Then trying something like the following (pseudo code)

```
old_wd = os.getcwd()
if workdir is not None:
    os.chdir(workdir)

# subprocess call

if workdir is not None:
    os.chdir(old_wd)

```

mcm:
@NicklasTegner I can do that, but it would have the same problem as mentioned in #251 so I'm not sure it really makes sense to incorporate into pypandoc. But if you just want to see if it solves the appveyor issue I can give that a try.

NicklasTegner:
yes please. It's just to see, if it's just changing dirs in general, or the cwd argument to subprocess that's at fault here :)

mcm:
Different failure, but also seemingly unrelated.

NicklasTegner:
@mcm  it looks like it was indeed the cwd argument that was at fault here.
The new failure was because it could reach the url to download the LaTeX engine, wich was unexpected. A simple restart of the build fixed that.

So this should be ready to merge

PR #254, opened 2022-03-18
Don't install examples

CyberTailor:
See Gentoo Bugzilla: https://bugs.gentoo.org/835475

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,7 +148,7 @@ output = pypandoc.convert_text(
     'md', format='html',
     extra_args=['--atx-headers'])
 # output == '# Primary Heading\r\n'
-output = pypandoc.convert(
+output = pypandoc.convert_text(
     '# Primary Heading',
     'html', format='md',
     extra_args=['--base-header-level=2'])
@@ -173,9 +173,6 @@ Please pass any filters in as a list and not as a string.
 
 Please refer to `pandoc -h` and the
 [official documentation](https://pandoc.org/MANUAL.html) for further details.
-
-> Note: the old way of using `convert(input, output)` is deprecated as in some cases it wasn't
-possible to determine whether the input should be used as a filename or as text.
 
 ## Dealing with Formatting Arguments
 
--- a/examples/services.py
+++ b/examples/services.py
@@ -43,7 +43,7 @@ class PandocPDFService(BasePandocService):
             '-o', self.file_object.name
         )
         # generate it using pandoc
-        self.service.convert(html, to_format, format=from_format, extra_args=extra_args)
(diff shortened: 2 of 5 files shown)


mgorny:
I think the author meant to include them in sdist. Poetry has some awful syntax for that, something like:

```
packages = {
  ...,
  { include = "examples/*", format = "sdist" },
}
```

and so on. LICENSE and README can be omitted as they're picked up by the default rules.

NicklasTegner:
hi @CyberTailor  thanks for the contributions.

Indeed as @mgorny  pointed out, it was expected behavior to include the examples.

Please let me know what issue you are having, as I couldn't find enough information from the link you provided :)

CyberTailor:
> Indeed as @mgorny pointed out, it was expected behavior to include the examples.

MANIFEST.in already does that.

> Please let me know what issue you are having

See build log of linked bug:

```
 * ERROR: dev-python/pypandoc-1.7.2::guru failed (install phase):
 *   Package installs 'examples' package which is forbidden and likely a bug in the build system.
```

NicklasTegner:
Ohh okay. I see @CyberTailor 
Would would be a prefered solution?
Removing the examples out right, or doing what @mgorny  suggested and only include them in sdist?

sthagen:
I cannot imagine any distribution removing examples from a source package - suggest to follow the route proposed by @mgorny

NicklasTegner:
@sthagen  maybe I'm just misreading it. But what are you suggesting?

mgorny:
> I cannot imagine any distribution removing examples from a source package - suggest to follow the route proposed by @mgorny

For the record, I think many Python projects nowadays assume that sdists are only to be consumed by pip (or pip-alike package managers) and should contain only files strictly needed to install the package in a venv, often skipping docs, examples and tests to reduce sdist size. Not that I liked it but at this point I'm pretty much assuming sdists are incomplete and going straight for git snapshots when packaging anything for Gentoo.

sthagen:
@NicklasTegner I referred to the proposal of @mgorny: 
> I think the author meant to include them in sdist. Poetry has some awful syntax for that, something like:
> 
> ```
> packages = {
>   ...,
>   { include = "examples/*", format = "sdist" },
> }
> ```
> 
> and so on. LICENSE and README can be omitted as they're picked up by the default rules.

That would mean add stanzas like the following in the `pyproject.toml` file:
```
 { include = "examples/*", format = "sdist" },
 { include = "tests/*", format = "sdist" },
```

Otherwise, I also gave up on `sdist` packages like @mgorny stated always hunting for the source repository to execute tests and learn from examples.

Issue #260, opened 2022-03-23
Python 2.7 no longer supported


scottclowe:
The code base no longer supports Python 2.7.

In of itself, that's not such a bad thing because the sunset of Python 2.7 was two years ago. It's fine to drop support for Python 2.7 at this point.

However, it is a bit of an issue that the current release of pypandoc claims to support Python 2.7 when it does not. If pypandoc is not intended to support Python 2.7 any more, it needs to have the Python 2.7 Trove classifier removed [[a]](https://github.com/NicklasTegner/pypandoc/blob/02f94233d3d3d0bf3c4d80e11b858e1cd22a4c0f/setup.py#L101) [[b]](https://github.com/NicklasTegner/pypandoc/blob/02f94233d3d3d0bf3c4d80e11b858e1cd22a4c0f/pyproject.toml#L19-L20), and have `python_requires=">=3.6"` set correctly in setup.py (not currently set) and [pyproject.toml](https://github.com/NicklasTegner/pypandoc/blob/02f94233d3d3d0bf3c4d80e11b858e1cd22a4c0f/pyproject.toml#L44). This is important so that pip knows which version of pypandoc it can install on any given version of python, and users installing pypandoc on an old version of python will install a valid version of pypandoc that can run on their system.

The changes which were made that are incompatible with Python 2.7 are:
- The use of `FileNotFoundError` (requires python>3), introduced in v1.5 by commit ffe8c52
  https://github.com/NicklasTegner/pypandoc/blob/02f94233d3d3d0bf3c4d80e11b858e1cd22a4c0f/pypandoc/pandoc_download.py#L106
- Use of the `typing` module (requires python>=3.5), introduced by commit 2cfcc1e (unreleased)
  https://github.com/NicklasTegner/pypandoc/blob/02f94233d3d3d0bf3c4d80e11b858e1cd22a4c0f/pypandoc/pandoc_download.py#L12
  https://github.com/NicklasTegner/pypandoc/blob/02f94233d3d3d0bf3c4d80e11b858e1cd22a4c0f/pypandoc/__init__.py#L3-L4

I've gone through the history of pypandoc releases, and the most recent one to run the affected code correctly on Python 2.7 was v1.3.3 (I tested to find this by trying to run the command `python -c "import pypandoc; pypandoc.download_pandoc(targetfolder='$PWD/test')"` on Linux).

The bleeding edge version does not install on Python 2.7, with a syntax error caused by type hinting:
```
[...]

PR #262, opened 2022-04-03
BUG+REL: Restore Python2.7 support

scottclowe:
Restores Python 2.7 support for a 1.7.5 hotfix before Python 2.7 support is dropped.

Tested on Python 2.7.18 on Ubuntu 20.04.4 with the following shell code:
```
sudo apt-get remove pandoc
rm -rf test
mkdir test
python
```
```python
import os
import pypandoc
dirname = os.path.join(os.getcwd(), "test")
pypandoc.download_pandoc(targetfolder=dirname)
os.environ["PATH"] += os.pathsep + dirname
pypandoc.convert_file("test.md", "rst", outputfile="test/test.rst")
pypandoc.convert_file("README.md", "rst", outputfile="test/README.rst")
exit
```
```sh
PATH="${PATH}:${PWD}/test"
python tests.py
```

Once merged, the py27support branch needs to be built as a wheel and released as 1.7.5 on PyPI.

### DIFF
--- a/pypandoc/__init__.py
+++ b/pypandoc/__init__.py
@@ -15,7 +15,7 @@ from .pandoc_download import DEFAULT_TARGET_FOLDER, download_pandoc
 from .py3compat import cast_bytes, cast_unicode, string_types, url2path, urlparse
 
 __author__ = u'Juho Vepsäläinen'
-__version__ = '1.7.4'
+__version__ = '1.7.5'
 __license__ = 'MIT'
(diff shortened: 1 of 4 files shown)


scottclowe:
Addresses #260.

NicklasTegner:
@scottclowe  this is now published to v1.7.5

Sidenote: Should probably make a cd script that runs
python setup.py download_pandoc bdist_wheel sdist on all platforms automatically when tagged.

PR #272, opened 2022-05-05
Handle pandoc logging returning two unknown log levels

### LINKED ISSUE #269: KeyError in _classify_pandoc_logging when the message format does not fit the assumptions
Since #250, I have an issue with the _classify_pandoc_logging function : when the parsed message does not fit the expected format, the parsing can end in a KeyError. For example, [pandoc-include](https://github.com/DCsunset/pandoc-include) has a warning formatted like this: 

`[Warn] included file not found: path_of_the_file`

`Warn` isn't in the dictionnary of expected log levels, which raises a KeyError on line 422: 
```python
yield level_map[level], "\n".join(log_msgs)
```

H0R5E:
This is a follow up to #271, which deals with the case where pandoc produces more than one logging message with non-standard levels, followed by another log message. For example, the following pandoc output still produced a `KeyError`:

```
[WARN] This is some message on
two lines
[ERR] This is a second message.
[ERROR] This is a third message.
```

This PR updates the `_classify_pandoc_logging` function and the `test_classify_pandoc_logging_invalid_level` test to correctly handle this case.

I have also updated the `level_map` dictionary which converts the pandoc levels to python codes, so that it only contains levels [formally used by pandoc](https://github.com/jgm/pandoc/blob/5e1249481b2e3fc27e845245a0c96c3687a23c3d/src/Text/Pandoc/Logging.hs#L44).

Fixes #269

### DIFF
--- a/pypandoc/__init__.py
+++ b/pypandoc/__init__.py
@@ -396,14 +396,14 @@ def _classify_pandoc_logging(raw, default_level="WARNING"):
     # that does not conform to the pandoc standard, use the default_level 
     # value instead.
     
+    # Available pandoc logging levels adapted from:
+    # https://github.com/jgm/pandoc/blob/5e1249481b2e3fc27e845245a0c96c3687a23c3d/src/Text/Pandoc/Logging.hs#L44
     def get_python_level(pandoc_level):
         
-        level_map = {"CRITICAL": 50,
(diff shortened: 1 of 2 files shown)


NicklasTegner:
Fantastic work.

PR #281, opened 2022-05-22
Making a pypandoc_binary version with included platform specific Pandoc binaries

### LINKED ISSUE #279: Creating a pypandoc-binary
I have been using pypandoc for a while now, and I always create a .whl with pandoc present in it as it's much easier to use.

It would be great if this package can hev:
 - pypandoc
 - pypandoc-binary
packages in pypi.

Somewhat like how the pysqlite3 project does it:
https://pypi.org/project/pysqlite3-binary/
https://pypi.org/project/pysqlite3/

That would help a lot for folks who just want to get a working version of pandoc directly without mucking around with system level dependencies.

NicklasTegner:
fixes #279


AbdealiJK:
This looks great !
Is the plan to do a pypi release now also ?

PR #284, opened 2022-06-13
Parse lua filters in filters argument

domvwt:
Small change to correctly parse lua filters passed to the `filters` argument.

### DIFF
--- a/tests.py
+++ b/tests.py
@@ -307,6 +307,33 @@ class TestPypandoc(unittest.TestCase):
         found = re.search(r'10.1038', written)
         self.assertTrue(found is None)
 
+    def test_conversion_with_python_filter(self):
+        markdown_source = "**Here comes the content.**"
+        python_source = '''\
+        #!/usr/bin/env python
+
+        """
+        Pandoc filter to convert all regular text to uppercase.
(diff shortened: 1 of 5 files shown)


domvwt:
I've added some tests to check that filter ordering is preserved!

tests.py
```diff
@@ -306,6 +307,89 @@ def test_conversion_with_empty_filter(self):
...
+            toJSONFilter(caps)
+        """
+        python_source = textwrap.dedent(python_source)
+        with closed_tempfile(".lua", lua_source) as tempfile1:
+            with closed_tempfile(".py", python_source) as tempfile2:
+                output = pypandoc.convert_text(
+                    markdown_source, to='html', format='md', outputfile=None, filters=[tempfile1, tempfile2, tempfile1]
+                ).strip()
+                expected = "<p>2 3 4 5</p>"
+                self.assertTrue(output == expected)
```
NicklasTegner, line 386: We also want an assert here, that check if the filters are in the same order, as they are supplied

domvwt: These results should have only been produced by the correct ordering but I've changed the setup to make it explicit.

PR #296, opened 2022-09-20
Issue295 fixes #295

### LINKED ISSUE #295: list index out of range in `_get_pandoc_urls`
```python
  File "/tmp/runner/build-docs/lib/python3.8/site-packages/sphinx/config.py", line 347, in eval_config_file
    exec(code, namespace)
  File "/home/runner/work/<org>/<repo>/docs/source/conf.py", line 37, in <module>
    download_pandoc(
  File "/tmp/runner/build-docs/lib/python3.8/site-packages/pypandoc/pandoc_download.py", line 228, in download_pandoc
    pandoc_urls, _ = _get_pandoc_urls(version)
  File "/tmp/runner/build-docs/lib/python3.8/site-packages/pypandoc/pandoc_download.py", line 57, in _get_pandoc_urls
    version = pandoc_urls_list[0].split('/')[5]
IndexError: list index out of range
```

Seems like the logic to load the `releases/` page has an empty list of URLs since it takes some time to load the URLs. This needs to be fixed ASAP as it breaks the use of `pypandoc` in all existing pipelines.

JessicaTegner:
This issue fixes a critical error, in download_pandoc, where the urls to the binaries stopped being populated.

This is a suggested solution, that I'll leave open for discussion before merging.

### DIFF
--- a/pypandoc/pandoc_download.py
+++ b/pypandoc/pandoc_download.py
@@ -1,6 +1,5 @@
 # -*- coding: utf-8 -*-
 
-import json
 import logging
 import os
 import os.path
@@ -12,6 +11,7 @@ import sys
 import tempfile
 from typing import Union
 
+import urllib
 try:
     from urllib.request import urlopen
 except ImportError:
@@ -44,23 +44,27 @@ def _get_pandoc_urls(version="latest"):
     :return: str version: actual pandoc version. (e.g. "lastest" will be resolved to the actual one)
     """
     # url to pandoc download page
-    api_url = "https://api.github.com/repos/jgm/pandoc/releases/" + \
-          ("tags/" if version != "latest" else "") + version
+    url = "https://github.com/jgm/pandoc/releases/" + \
+          ("tag/" if version != "latest" else "") + version
+    # try to open the url
+    try:
+        response = urlopen(url)
+        content = response.read()
+        pattern = re.compile(r"pandoc\s*([\d.]+)")
+        version = re.search(pattern, content.decode("utf-8")).group(1)
+    except urllib.error.HTTPError as e:
+        raise RuntimeError("Invalid pandoc version {}.".format(version))
+        return
     # read the HTML content
-    response = urlopen(api_url)
+    response = urlopen("https://github.com/jgm/pandoc/releases/expanded_assets/"+version)
     content = response.read()
-    api_json = json.loads(content)
-    release_assets = {}
-    for asset in api_json["assets"]:
-        release_assets[asset["name"]] = asset["browser_download_url"]
     # regex for the binaries
     processor_architecture = "arm" if platform.uname()[4].startswith("arm") else "amd"
-    regex = re.compile(r"pandoc\-.*(?:"+processor_architecture+"|x86|mac).*\.(?:msi|deb|pkg)")
-    matched_keys = {key for key, value in release_assets.items() if regex.match(key)}
+    regex = re.compile(r"/jgm/pandoc/releases/download/.*(?:"+processor_architecture+"|x86|mac).*\.(?:msi|deb|pkg)")
     # a list of urls to the binaries
-    pandoc_urls_list = [release_assets[key] for key in matched_keys]
+    pandoc_urls_list = regex.findall(content.decode("utf-8"))
     # actual pandoc version
-    version = pandoc_urls_list[0].split('/')[7]
+    version = pandoc_urls_list[0].split('/')[5]
     # dict that lookup the platform from binary extension
     ext2platform = {
         'msi': 'win32',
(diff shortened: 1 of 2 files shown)


kevalmorabia97:
Thanks for the fix!

I always see another error when running the pypandoc download:
```python
 [INFO] Downloading pandoc from https://github.com/jgm/pandoc/releases/download/2.19.2/pandoc-2.19.2-1-amd64.deb ...
[INFO] Unpacking /tmp/tmpyhpyagw4/pandoc-2.19.2-1-amd64.deb to tempfolder...
[INFO] Copying pandoc to /tmp/runner/build-docs/bin ...
[INFO] Making /tmp/runner/build-docs/bin/pandoc executeable...
[INFO] Copying pandoc-citeproc to /tmp/runner/build-docs/bin ...
Error:  Didn't copy pandoc-citeproc
```
This is just a print statement and does the cause the program to crash but still is a bit annoying. Is this something that can be easily disabled?

JessicaTegner:
@kevalmorabia97  this can be disabled indeed. See the README for instructions, specifically under the "logging messages" section.

kevalmorabia97 (approved):
Thanks for the quick fix! I tested it on MacOS and it works.

FedeMPouzols:
I came across this issue, thanks for the quick fix! I can confirm it works well on Ubuntu 22.04.

Querying api.github.com makes perfect sense to me, and will hopefully be more stable.

JessicaTegner:
@FedeMPouzols  good that it also works there.

Issue that I don't like, is that api.github.com has a ratelimiting of 60 requests / hour for non-authenticated users.

This might not seem like a big deal, but consider that before it was unlimite due to reading the html, so bigger pipelines might break if it's ran many times an hour (even our own tests, breaks sometime because of it).

I have not yet found a fix for this, that I like.
I do take suggestions.

kevalmorabia97:
@JessicaTegner how about we use https://github.com/jgm/pandoc/releases/expanded_assets/2.19.2 urls? I think it might not work for `latest` version in which case we would need to fall back to what you have now. What do you think?

JessicaTegner:
@kevalmorabia97  that would solve the rate limiting issue for every version except if using "latest".
Since we are only hitting api.github.com once, if we use "latest" as version, the rate limiting will still happen.

Reason for this, is that we are extracting the asset html download urls from the json data returned from the call.

If we could find another way of getting the latest version number, that didn't involve hitting api.github.com, your solution might just work perfectly.

kevalmorabia97:
What if we cache the latest version and let the cache be alive for a minute, this would ensure we hit the api almost once per minute

JessicaTegner:
that could also work, however I think I have found a solution between what you suggested and what I have found after doing some research. Do let me know what you think :)

kevalmorabia97 (approved):
This is even better! I tested it in MacOS and it works

JessicaTegner:
@kevalmorabia97   also works on Windows here.

I don't like that the tests are failing, but that seems to be for other reasons, not a cause of this pr.


I'll leave this open for a little, to see if other people has comments, while I look at better CI pipeline options.

kevalmorabia97:
Sounds good. For now I'm passing a hard coded url to the pandoc_download function

enochtangg:
Is it possible to make a new release/tag for this working state?

pypandoc/pandoc_download.py
```diff
@@ -45,8 +46,17 @@ def _get_pandoc_urls(version="latest"):
     # url to pandoc download page
     url = "https://github.com/jgm/pandoc/releases/" + \
           ("tag/" if version != "latest" else "") + version
+    # try to open the url
+    try:
+        response = urlopen(url)
+        content = response.read()
+        pattern = re.compile(r"pandoc\s*([\d.]+)")
+        version = re.search(pattern, content.decode("utf-8")).group(1)
```
davidbgk, line 54: It might be easier/faster to parse the returned URL here:

```py
>>> from urllib.request import urlopen
>>> url = "https://github.com/jgm/pandoc/releases/latest"
>>> response = urlopen(url)
>>> response.url
'https://github.com/jgm/pandoc/releases/tag/2.19.2'
```

And once you have that URL, you can `.replace('tag', 'expanded_assets')` to get the one with links.

davidbgk:
@enochtangg as a temporary option, you can update your dependencies to `pypandoc @ https://github.com/JessicaTegner/pypandoc/archive/refs/heads/master.zip` to try that it works on your environment.

JessicaTegner:
@enochtangg  that has now been done.

[PyPandoc v1.9](https://pypi.org/project/pypandoc/1.9/) - [pypandoc-binary v1.9](https://pypi.org/project/pypandoc-binary/1.9/)

PR #301, opened 2022-09-28
Sort files before proc fixes #292

### LINKED ISSUE #292: Sorting the order of multiple globbed files
PR https://github.com/JessicaTegner/pypandoc/issues/248 supports the ability to pass multiple files, but I notice that if I pass a glob pattern (eg `*.html`) to combine multiple files into a single PDF, for example:

- in the command line version of pandoc, the docs are passed in alphabetical sort order;
- in pypandoc, the files are parsed in an arbitrary order.

If a glob pattern is passed, eg to capture files `01.md`, `02.md` etc, it would be really useful if the sort order were respected.

hey24sheep:
Added sort files before proccessing, fixes issue #292

### DIFF
--- a/pypandoc/__init__.py
+++ b/pypandoc/__init__.py
@@ -337,6 +337,8 @@ def _convert_input(source, format, input_type, to, extra_args=(),
             input_file = source
     else:
         input_file = []
+    
+    input_file = sorted(input_file)
     args = [__pandoc_path, '--from=' + format]
 
     args.append('--to=' + to)


hey24sheep:
I think hacktoberfest will count PR from October. So please do not merge right now

JessicaTegner:
Hi @hey24sheep  fantastic. I wont merge until then. But please do pull the latest changes in from master, so the tests and so on will run :)

hey24sheep:
Sure, I will

PR #311, opened 2022-10-28
Updated pandoc-citeproc handling after downloading

JessicaTegner:
This pr changes the handling of pandoc-citeproc, to check if it's Included.If it is, copy like normal.
If not, do nothing


JessicaTegner:
@AbdealiJK let me know how this works for ya :)

AbdealiJK (approved):

AbdealiJK:
Makes sense :+1:

PR #312, opened 2022-10-28
Remove logic for python2

AbdealiJK:
The library currently cannot be installed on py2 because of `setup_requires=">=3.6"`
So, remove code for python 2

### DIFF
--- a/pypandoc/__init__.py
+++ b/pypandoc/__init__.py
@@ -13,10 +13,11 @@ import textwrap
 import glob
 from pathlib import Path
 from urllib.parse import urlparse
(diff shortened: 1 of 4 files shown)


pypandoc/__init__.py
```diff
@@ -1,5 +1,4 @@
 # -*- coding: utf-8 -*-
```
funkyfuture, line 1: these headers are also obsolete.

pypandoc/py3compat.py
funkyfuture, line 1: could the remaining contents of this module moved into another one?

PR #313, opened 2022-10-28
ci: Fix conditions for building and publishing

AbdealiJK:
Based ont he github documentation:

github.events.ref: Is the ref of a release that has been created github.ref: If the branch that is being used (depending on the action)
 - create a release:      It will be refs/tags/<tag_name>
 - create a pull request: It will be refs/pull/<pr_number>/merge
 - push to a branch:      It will be refs/heads/<branch_name>

For publishing to pypi, we are targeting the "create a release" event. In which case:
 - github.events.ref will be refs/tags/<tag_name>
 - github.ref will also be refs/tags/<tag_name>

So, it looks like our IF condition was incorrect.

Ref: https://docs.github.com/en/actions/learn-github-actions/environment-variables

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -72,7 +72,7 @@ jobs:
 
   builder_pypandoc_binary:
     needs: [test]
-    if: github.ref == 'refs/heads/master'
+    if: github.ref == 'refs/heads/master' || startsWith(github.event.ref, 'refs/tags/v')
     strategy:
       matrix:
         # Ref: https://cibuildwheel.readthedocs.io/en/stable/options/#archs
@@ -103,7 +103,7 @@ jobs:
 
   publisher_release:
     needs: [builder_pypandoc, builder_pypandoc_binary]
-    if: startsWith(github.event.ref, 'refs/tags/v') && github.ref == 'refs/heads/master'
+    if: startsWith(github.event.ref, 'refs/tags/v')
     runs-on: ubuntu-latest
     steps:
       - name: Check out repository
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -72,7 +72,7 @@ jobs:
 
   builder_pypandoc_binary:
     needs: [test]
-    if: github.ref == 'refs/heads/master' || startsWith(github.event.ref, 'refs/tags/v')
+    if: github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/tags/v')
     strategy:
       matrix:
         # Ref: https://cibuildwheel.readthedocs.io/en/stable/options/#archs
@@ -103,7 +103,7 @@ jobs:
 
   publisher_release:
     needs: [builder_pypandoc, builder_pypandoc_binary]
-    if: startsWith(github.event.ref, 'refs/tags/v')
+    if: startsWith(github.ref, 'refs/tags/v')
     runs-on: ubuntu-latest
     steps:
       - name: Check out repository


JessicaTegner:
@AbdealiJK  this seems very strange.

* Building and publishing our latest version, should happen on the master branch, regardless if a tag is specified or not.
* If a tag is specified that starts with "v" and we are on a master branch, we should do a release to pypi.

From what I can see from your suggested changes, this does not seem to be the case anymore?

AbdealiJK:
I don't think github actions has a way of detecting `a tag is specified that starts with "v" and we are on a master branch` (atleast not with the variables we are using right now.

From what I read in the documentation ...

The variables we are using are:

`github.ref` [doc](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context)
> The fully-formed ref of the branch or tag that triggered the workflow run. For workflows triggered by push, this is the branch or tag ref that was pushed. For workflows triggered by pull_request, this is the pull request merge branch. For workflows triggered by release, this is the release tag created. For other triggers, this is the branch or tag ref that triggered the workflow run. This is only set if a branch or tag is available for the event type. The ref given is fully-formed, meaning that for branches the format is refs/heads/<branch_name>, for pull requests it is refs/pull/<pr_number>/merge, and for tags it is refs/tags/<tag_name>. For example, refs/heads/feature-branch-1.

`github.event` [doc](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context)
> The full event webhook payload. You can access individual properties of the event using this context. This object is identical to the webhook payload of the event that triggered the workflow run, and is different for each event. The webhooks for each GitHub Actions event is linked in "[Events that trigger workflows](https://docs.github.com/en/articles/events-that-trigger-workflows)." For example, for a workflow run triggered by the [push event](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#push), this object contains the contents of the [push webhook payload](https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#push).

and the 2 events we enabled in the yaml file are:
  - [push](https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#push)
  - [pull_request](https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#pull_request)

Leaving pull request aside for now ...

Push can have 2 cases:
 - Push a tag
 - Push a branch

AbdealiJK:
When I am reading through the documentation again, I am realizing that `github.ref` and `github.event.ref` will always be the same for the `push` event.

So, tweaking your logic:
 - Building and publishing our latest version, should happen on **any push to** the master branch, regardless if a tag is specified or not.
 - If a tag is specified that starts with "v" **and pushed**, we should do a release to pypi.

JessicaTegner:
yes I guess that's the best we can do.
It's just strange, cause the current logic did work on v1.9 🤔

AbdealiJK:
I believe the issue was introduced in where the IF conditions for both PR and Publish cases were being written
https://github.com/JessicaTegner/pypandoc/commit/ad520d1bca40f6bf1ad3d77f357068c326e8c45c#diff-944291df2c9c06359d37cc8833d182d705c9e8c3108e7cfe132d61a06e9133ddR73

In v1.9 from what I see, there were 2 separate yaml files being used each with their own `on:` criteria - which was merged in https://github.com/JessicaTegner/pypandoc/commit/7ccd216e0949e230fa8e69c65ed42c143185d6a3

JessicaTegner:
Do you think the current implementation you've written here, will work?

AbdealiJK:
I believe so
But I am not an expert at GitHub actions. So the only way to know is to try it out :)

JessicaTegner:
well here we go: https://github.com/JessicaTegner/pypandoc/actions/runs/3347112634

PR #314, opened 2022-10-28
ci: Add condition for the both publish steps

AbdealiJK:


AbdealiJK:
@JessicaTegner Looks like I forgot to fix the condition for the source distribution :|
Added it here

JessicaTegner:
nicely spotted. Let's hope it works this time
https://github.com/JessicaTegner/pypandoc/actions/runs/3347340476

PR #327, opened 2023-03-06
Add release.md to MANIFEST.in

hugobuddel:
This should fix `test_basic_conversion_from_file_pattern` and `test_basic_conversion_from_file_pattern_with_input_list` when ran on the released code.

These tests check for the presence of the string `making a release` which is taken from `release.md`. Therefore, `release.md` must be in the release tarball.

See also https://github.com/JessicaTegner/pypandoc/pull/259/files#r1126859693

However, this PR has not been tested.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -2,4 +2,5 @@ include README.md
 include LICENSE
 include tests.py
 include filter_test.md
+include release.md
 include examples/*


JessicaTegner:
I would say so. However, looking at the 2 (at the moment released) packages, it looks like only the tarball has a *.md file in it(the readme).

But yes, I'm in favor of using what's already there (or, creating temp files, as the testing already has functions to handle such things).

Thinking about it more, it would probably be best to just use temp files, cause that would make it all independent

hugobuddel:
Related, it would be 'better' to test the build package rather than the source repository. In my own projects I try to enforce this by putting the package (which would be the `pypandoc` directory) in a `src` directory and the tests in a separate `tests` directory. That way the tests can never accidentally import the source directory, but rather the built and installed package. Then this error would have been caught in the CI, but not sure whether that change is worth the effort.

Is it OK to leave it to you how to proceed? That is, choose one or more of

* Merging this as-is, which would probably fix  the immediate problem.
* Use true test files in a temporary directory.
* Test the 'release' package rather than the source.

My immediate problem was installing the app 'apostrophe' on Guix, which requires pypandoc, which failed to build, see https://issues.guix.gnu.org/62013 . Writing this PR was already a rabbit hole deeper than I intended, so just merging this as-is would be sufficient for me for now. In the meantime I've made a patch to update Guix to use pypandoc 1.7.5, which does work.

JessicaTegner:
Seems resonable @hugobuddel. Kind of related, does the patches in #328  also work for you. Then if that's the case, I think we are gonna move to that one, due to it using actual temp test files. THen in another later pr, I'll seperate the code into an src dir, like you suggested

hugobuddel:
> Seems resonable @hugobuddel. Kind of related, does the patches in #328 also work for you. Then if that's the case, I think we are gonna move to that one, due to it using actual temp test files. 

I don't think #328 is related, it touches a different test

JessicaTegner:
Cool. I think what we'll do is use temporary markdown file, per your secondary suggestions. Then I'll work on refactoring the testing

PR #328, opened 2023-03-07
Tests fixes

keszybz:


JessicaTegner:
Can you please describe what is getting done here. Seems like the work of #327

keszybz:
> Can you please describe what is getting done here. Seems like the work of #327

I don't think those are related. The first patch in this PR fixes an obvious bug in tests. I think the commit description is very clear. The other two patches are minor cleanups. ??

JessicaTegner:
Fantastic work :)

Issue #333, opened 2023-05-18
Update the Dockerfile to use Python 3


DisSupEng:
Hi,

I have noticed that the Dockerfile is quite out of date and uses a Python 2 image. I was wondering if you would be open to me updating this to use Python 3?

Kind Regards

PR #334, opened 2023-05-23
Update Dockerfile to use Python 3.9

DisSupEng:
# Changes
- Updated `Dockerfile` to use `python:3.9` image
- Added `docker-compose.yml` file to easily set up development environment
- Updated `REAME` with my contribution

Addresses issue #333 

**Note: when running flake8 I get a lot of formatting errors but I haven't changed the code so I'm not sure if this is a pre-existing issue**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@
 #     docker build -t <user>/pypandoc .
 # --------------------------------------------------------------------------
 
-# Switch to a more recent version of Python
+# Use Python 3.9
 FROM  python:3.9
 # Not sure if this line needs updating, I think this person no longer works on the project?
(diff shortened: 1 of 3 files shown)


DisSupEng:
Hi @JessicaTegner thank you for running those tests. I can see that there is an error for cryptography when using pypy3.7.

This is the commit the exception was added in: https://github.com/pyca/cryptography/commit/60d3f709eb96ba5df9bdfede5c024c1f2db9f0e5

It looks to have been added three months ago. I could try pinning cryptography to a previous version but I'm assuming they don't want to use 3.7 for a reason so that might not be the best idea.

JessicaTegner:
@DisSupEng  I would say to pin cryptography only on that specific pypy version, as that's the only one failing. I don't know if this is possible though

PR #340, opened 2023-08-30
make meaningless encoding parameter from convert_file method deprecated

chrisdecker1201:
I think this encoding parameter is a relic of copying the `convert_text` method, because it's not used.

As far as I understood the current behavior of the `convert_file` method, it will just execute pandoc with the given file(s).

I think there might be the possibility to implement the `encoding` parameter in two ways:

1. check the encoding and bring an error if it's the wrong encoding
2. convert the given encoding to utf-8 in a temporary file and use this temporary file as input for pandoc

Nevertheless, I think to remove confusion it might be the best solution to just remove the parameter :)

### DIFF
--- a/pypandoc/__init__.py
+++ b/pypandoc/__init__.py
@@ -116,7 +116,7 @@ def convert_file(source_file:Union[list, str, Path, Generator], to:str, format:U
     :param list extra_args: extra arguments (list of strings) to be passed to pandoc
             (Default value = ())
 
-    :param str encoding: the encoding of the file or the input bytes (Default value = 'utf-8')
+    :param str encoding (deprecated): the encoding of the input bytes (Default value = 'utf-8')
 
     :param str outputfile: output will be written to outputfile or the converted content
             returned if None. The output filename can be specified as a string
@@ -137,6 +137,10 @@ def convert_file(source_file:Union[list, str, Path, Generator], to:str, format:U
     :raises OSError: if pandoc is not found; make sure it has been installed and is available at
             path.
     """
+    # TODO: remove 'encoding' parameter and warning
+    if encoding != "utf-8":
+        logger.warning("The 'encoding' parameter will be removed in version 1.13. Just remove the parameter, because currently the method does not use it.")
+
     # This if block effectively adds support for pathlib.Path objects
     # and generators produced by pathlib.Path().glob().
     if not isinstance(source_file, str):


JessicaTegner:
@chrisdecker1201  I agree. We could remove it.
TO do this, I would suggest having it in the function signature as is (basically current behaviour), but issue a deprecation warning that it will be removed in pypandoc 1.13 so 1.11 (current) 1.12 (warning) 1.13 (removed).

Let me know what you think

chrisdecker1201:
@JessicaTegner I think this is a good idea. But I've no idea how to set this property to deprecated.

JessicaTegner:
@chrisdecker1201  it's not that complicated, it is the following:

* Change no functional code.
* Let Write a deprecated in the docstring like the following

    :param str encoding (deprecated): the encoding of the input bytes (Default value = 'utf-8')

* Use the logger to print a warning message.

chrisdecker1201:
@JessicaTegner Like that?

Thank you for taking your time for this 🙂

pypandoc/__init__.py
```diff
@@ -137,6 +137,10 @@ def convert_file(source_file:Union[list, str, Path, Generator], to:str, format:U
     :raises OSError: if pandoc is not found; make sure it has been installed and is available at
             path.
     """
+    # TODO: remove 'encoding' parameter and warning
+    if encoding:
```
JessicaTegner, line 141: encoding will always be evaluated to true, because the default value is "utf-8". I suggest just doing

if encoding != "utf-8":

pypandoc/__init__.py
```diff
@@ -137,6 +137,10 @@ def convert_file(source_file:Union[list, str, Path, Generator], to:str, format:U
...
+        logger.warning("The 'encoding' parameter will be removed in version 1.13. Just remove the parameter, because it has no effect anyway.")
```
JessicaTegner, line 142: Explain quickly why it doesn't have any effect. E.g: I does no longer have an effect, because it was used for python 2 compatibility

chrisdecker1201:
@JessicaTegner Just question: Do you just remove the parameter then in version 1.13 or should I prepare a PR?

JessicaTegner:
@chrisdecker1201  there is really 2 options.

* You are more than welcome to make a pr for it
* Or alternatively, you are more than welcome to make an issue descriping it, so someone else can take it on. It might even be a good first issue for some

Issue #342, opened 2023-09-05
source verification runs before cworkdir


Lrakotoson:
Hi 👋  

In the `convert_file(...)` function, the `cworkdir` is only aknowledged in `_convert_input(...)` function, which is called after the source verification at `_identify_path(source_file)`.

It would be more practical to check the files once the working directory has been defined.  
Otherwise, it would be necessary either to always give a path relative to the old working directory, or to give lists of absolute paths. Which is not really optimal when using extra_args.

JessicaTegner:
ahh I see.
Nice catch.
Will look into this soon, if someone hasn't already.

It's at the top of the priority list

PR #344, opened 2023-10-04
Change when we do source file verification

JessicaTegner:
Change when we do source file verification #342. Previously we did sourceverificiation before looking at the working directory. Now we convert all source files to an absolute file path if needed, then perform globbing and then do source verification. This fixes a bug where we would not find source files if the working directory was not the same as the source file directory.


JessicaTegner:
@Lrakotoson what you think of this solution? :)

Everyone else, feel free to come with suggestion or review

Issue #339, opened 2023-08-28
In Windows, the function _get_pandoc_version() dont receive the path correctly


excript:
In Windows, pandoc exists in directory, but a search for them occurs without the name of file, in this case, pandoc.exe.

The 567 line fails because the path is incorrect, no have file name pandoc.exe.

![image](https://github.com/JessicaTegner/pypandoc/assets/10388739/46998adb-3fbc-43ff-9e66-7d354084c686)

JessicaTegner:
This is an interesting one. THanks for pointing it out.

The steps that would need to be taken to fix this issue (mostly documenting it for myself for later, but others are welcome to take a stap at it) are fixing the "_ensure_pandoc_path", most likely adding ".exe" to all search paths (that are a folder like "file/to/pandoc" so it becomes "path/to/pandoc.exe", if we are running on Windows

PR #345, opened 2023-10-04
pass path to _get_pandoc_version() on windows correctly

JessicaTegner:
Fixed so In Windows, the function _get_pandoc_version() dont receive the path correctly #339


JessicaTegner:
@excript let me know if this fixes your issue

Issue #355, opened 2024-02-09
Can pypandoc_binary support linux/arm64, because download_pandoc() seems to?


paul-gauthier:
Thanks for a great tool! I really appreciate the attention to making it easy to automagically install the pandoc executable. This makes it really easy to build tools on top of pandoc without asking end users to do out of band manual installs.

It appears that `download_pandoc()` is able to install pandoc on linux/arm64. But the `pypandoc_binary` package fails to install in that environment. Is it possible to upgrade pypandoc_binary to directly support linux/arm64?

Here's a transcript of a linux/arm64 docker container showing the `pip install pypandoc_binary` failing and the `download_pandoc()` working:

```
$ docker run --platform linux/arm64 -it python:3.10 /bin/bash

root@d781f0bbe5ba:/# uname -m
aarch64

root@d781f0bbe5ba:/# pip install pypandoc
Collecting pypandoc
  Downloading pypandoc-1.12-py3-none-any.whl (20 kB)
Installing collected packages: pypandoc
Successfully installed pypandoc-1.12

root@d781f0bbe5ba:/# pip install pypandoc_binary
ERROR: Could not find a version that satisfies the requirement pypandoc_binary (from versions: none)
ERROR: No matching distribution found for pypandoc_binary

root@d781f0bbe5ba:/# python
Python 3.10.13 (main, Feb  1 2024, 05:13:02) [GCC 12.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.

>>> import pypandoc
>>> pypandoc.convert_text('# some title', 'rst', format='md')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.10/site-packages/pypandoc/__init__.py", line 92, in convert_text
    return _convert_input(source, format, 'string', to, extra_args=extra_args,
  File "/usr/local/lib/python3.10/site-packages/pypandoc/__init__.py", line 337, in _convert_input
    _ensure_pandoc_path()
  File "/usr/local/lib/python3.10/site-packages/pypandoc/__init__.py", line 763, in _ensure_pandoc_path
    raise OSError("No pandoc was found: either install pandoc and add it\n"
OSError: No pandoc was found: either install pandoc and add it
to your PATH or or call pypandoc.download_pandoc(...) or
install pypandoc wheels with included pandoc.

>>> pypandoc.download_pandoc()
>>> pypandoc.convert_text('# some title', 'rst', format='md')
[...]

PR #356, opened 2024-02-14
Updated CI to produce arm builds off pypandoc_binary

JessicaTegner:
This pr addresses #355 more specifically, updates the CI/CD pipeline to produce arm builds of pypandoc_binary on linux and MacOS.
This is important for pypandoc_binary specifically, because we bundle a version of pandoc, which of course is architecture specific to the platform.

In addition, we also update older Actions to newer versions, and subsequently update the CI script.


JessicaTegner:
Was confirmed to be working in issue #355

PR #365, opened 2024-05-23
Fix invalid escape sequence in regex

syusuke9999:
Updated the regex in pandoc_download.py to correctly escape the period character. This change resolves the SyntaxWarning triggered by an invalid escape sequence and ensures compatibility with future Python versions.

### DIFF
--- a/pypandoc/pandoc_download.py
+++ b/pypandoc/pandoc_download.py
@@ -58,7 +58,7 @@ def _get_pandoc_urls(version="latest"):
     # regex for the binaries
     uname = platform.uname()[4]
     processor_architecture = "arm" if uname.startswith("arm") or uname.startswith("aarch") else "amd"
-    regex = re.compile(r"/jgm/pandoc/releases/download/.*(?:"+processor_architecture+r"|x86|mac).*\.(?:msi|deb|pkg)")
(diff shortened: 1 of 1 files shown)


JessicaTegner:
@syusuke9999  it seems that this change breaks the download pandoc functionality in the CI pipeline. Could you take a look at this

syusuke9999:
Thanks for the feedback, @JessicaTegner. I've adjusted the regex to fix the CI download issues and verified that it now works as intended. Please review the changes at your convenience.

JessicaTegner:
@syusuke9999  seems good to me. Thank you so much for  the pr :)

jauderho:
I'm not sure if this is correct. 

When trying to use aider (which has pypandoc as a dependency) and python 3.12, I got the following error message:

```
/usr/lib/python3.12/site-packages/pypandoc/pandoc_download.py:61: SyntaxWarning: invalid escape sequence '\.'
  regex = re.compile(r"/jgm/pandoc/releases/download/.*(?:"+processor_architecture+"|x86|mac).*\.(?:msi|deb|pkg)")
```


It would seem that the correct fix is to properly escape, thusly:
`regex = re.compile(r"/jgm/pandoc/releases/download/.*(?:" + processor_architecture + r"|x86|mac).*\.(:?msi|deb|pkg)")`

PR #372, opened 2024-08-23
Collect pandoc urls from github api

nstrmx:


JessicaTegner:
Hey @nstrmx  thanks for the PR.

Can you take a look at the failing checks. Do you think it's a PR issue, or something else?

nstrmx:
I've limited collected urls to a host processor_architecture at pypandoc/pandoc_download.py:73. Guess I was wrong there, tests run on linux, thus only linux urls are collected. Fixed it (hopefully).

nstrmx:
Fixed it (hopefully).

JessicaTegner:
Interestingly @nstrmx  it seems that the single test failing, is actually because of a 403 rate limit.
What would you say we could do about that? Asking because Idon't know that much about GHs api, when used in this way.

nstrmx:
If we add support for the GITHUB_API_KEY as an environment variable to be used in requests, will that solve this?

JessicaTegner:
yes, that would be acceptable. But I feel like that if that key isn't set, we should fall back to the old behaviour.
Due to the nature of pypandoc, where it's used a lot in CI/CD workflows, I do think that if it's not set, we should fall back to old behaviour so stuff doesn't break

nstrmx:
Should work now.

JessicaTegner:
Great work. Lastly, since we now should have 2 download behaviors, we should probably make a test casefor each behavior

PR #381, opened 2024-11-19
allow non-latex intermediaries for pdf generation

### LINKED ISSUE #380: Allow non-LaTeX intermediaries for PDF generation
As of `pandoc` 2.8.0.1, released in November 2019, `pdf` is a valid output format. Since well before then, `LaTeX` was not the only intermediary format for creating PDFs; we currently force `latex` as the intermediary, which poses problems when trying to use a different intermediary format.

To fix this, we should be able to remove [this line](https://github.com/JessicaTegner/pypandoc/blob/8bc3c51ee58d6ba33f14f52ad7c7f6b6b91b0ea2/pypandoc/__init__.py#L358) and actually use `pdf` as the output format. I've tested this locally to use typst as an intermediary, which previously failed with the failure `pdf-engine typst is not compatible with output format latex` but now correctly invokes and passes through the pandoc arguments.

ashi-agrawal:
fixes #380

### DIFF
--- a/pypandoc/__init__.py
+++ b/pypandoc/__init__.py
@@ -339,8 +339,7 @@ def _validate_formats(format, to, outputfile):
         )
 
     if base_to_format == "pdf":
-        # pdf formats needs to actually have a to format of latex and a
-        # filename with an ending pf .pdf
+        # pdf formats need a filename with an ending of .pdf
         if isinstance(outputfile, str):
             if outputfile[-4:] != ".pdf":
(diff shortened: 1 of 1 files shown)


JessicaTegner:
Hi @ashi-agrawal 

Some CI is failing. Would you be able to take a look at that

ashi-agrawal:
Hey @JessicaTegner -- sorry for missing this. Looks like all checks are passing?

PR #382, opened 2024-11-19
Refactor Duplicate Code in _identify_path Function

zvaad:
• Refactored the _identify_path function to remove redundant code and improve readability.
• Consolidated logic for path resolution and network path checking to avoid duplication.
• Simplified handling of lists using Python's all() function for better clarity.
This change ensures cleaner and more maintainable code without altering functionality.


JessicaTegner:
hey @zvaad  thank you for the pr.
So we have some CI failing, so idially a refactor should not make any tests break. Could you take another look at it?

PR #384, opened 2024-12-08
Fix convert_file for Python 3.13

cjwatson:
As of https://github.com/python/cpython/pull/117589 (at least), `Path.glob` returns an `Iterator` rather than `Generator` (which inherits from `Iterator`).  `convert_file` doesn't need to care about this distinction; it can reasonably accept both.

This previously caused a test failure along these lines:

```
______________________________________________________ TestPypandoc.test_basic_conversion_from_file_pattern_pathlib_glob _______________________________________________________

self = <tests.TestPypandoc testMethod=test_basic_conversion_from_file_pattern_pathlib_glob>

    def test_basic_conversion_from_file_pattern_pathlib_glob(self):
        received_from_str_filename_input = pypandoc.convert_file("./*.md", 'html').lower()
>       received_from_path_filename_input = pypandoc.convert_file(Path(".").glob("*.md"), 'html').lower()

tests.py:654:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

source_file = <map object at 0x7f83952c9420>, to = 'html', format = None, extra_args = (), encoding = 'utf-8', outputfile = None, filters = None, verify_format = True
sandbox = False, cworkdir = '/home/cjwatson/src/python/pypandoc', sort_files = True
[...]
        if not _identify_path(discovered_source_files):
>           raise RuntimeError("source_file is not a valid path")
E           RuntimeError: source_file is not a valid path

pypandoc/__init__.py:201: RuntimeError
```

### DIFF
--- a/pypandoc/__init__.py
+++ b/pypandoc/__init__.py
@@ -1,8 +1,8 @@
 # -*- coding: utf-8 -*-
 from __future__ import absolute_import, print_function, with_statement
 from typing import Iterable
+from typing import Iterator
 from typing import Union
-from typing import Generator
 
 import os
 import re
@@ -97,7 +97,7 @@ def convert_text(source:str, to:str, format:str, extra_args:Iterable=(), encodin
                           cworkdir=cworkdir)
 
 
-def convert_file(source_file:Union[list, str, Path, Generator], to:str, format:Union[str, None]=None,
+def convert_file(source_file:Union[list, str, Path, Iterator], to:str, format:Union[str, None]=None,
                  extra_args:Iterable=(), encoding:str='utf-8', outputfile:Union[None, str, Path]=None,
                  filters:Union[Iterable, None]=None, verify_format:bool=True, sandbox:bool=False,
                  cworkdir:Union[str, None]=None, sort_files=True) -> str:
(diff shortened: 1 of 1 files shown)


JessicaTegner:
hey @cjwatson 

It seems some CI is failing. Could you take a look at this

cjwatson:
One of your indirect dependencies seems to have dropped PyPy 3.8 support.  What do you want to do about that?  Judging from [PyPy's release notes](https://doc.pypy.org/en/latest/index-of-release-notes.html) it seems as though they've dropped 3.8 support, so maybe the easiest way out would be to just drop PyPy 3.8 support here, but that's not a decision I feel I can take for you.  If you don't want to do that then I guess it would require some version pinning on PyPy 3.8, which would probably be a moving target.

JessicaTegner:
there we go @cjwatson 
I updated the CI workflow, dropping support for python 3.7, 3.8 and pyp3.8 ( #387  )

If you are all done with the fix, this can be merged

Issue #397, opened 2025-01-31
Add py.typed


Goldziher:
Hi,

This library doesn't have a py.typed file, and hence considered untyped by mypy and other checkers. 

Solution:

add a `py.typed` file as part of the package. This is a small thing. 

Here is the pertinent info: https://peps.python.org/pep-0561/

kurtmckee:
This is a duplicate of #369.

PR #398, opened 2025-01-31
chore: add py.typed

Goldziher:
This PR addresses #397, adding a py.typed to the package.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,6 +19,7 @@ setup(
     author = module.__author__.encode('utf8'),
     author_email = pypandoc.__author_email__,
     packages = ['pypandoc'],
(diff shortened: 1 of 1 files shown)


kurtmckee (changes requested):
1. There isn't currently a `py.typed` file in the source tree. Did you test that the project can be built after adding this change?

2. mypy isn't being run against the project in CI to confirm that the type annotations are coherent. Adding a `py.typed` file to claim that the project is typed isn't much of a claim without actually testing the type annotations.

I recommend modifying CI to run mypy against the project to confirm that the type annotations are coherent.

sakshamgoyal-01:
can this be revisited?

PR #401, opened 2025-02-27
Remove Python 2 compatibility code

kurtmckee:
Exactly what it says on the tin.

Because the compatibility code is removed, the LICENSE file no longer needs extra attribution.


JessicaTegner:
Great work!

